PR #22, opened 2015-08-19
Proper name case for Forward and Back bitmaps

nemerle:
Related to linux build.

I don't remember if '/' path separator works properly on windows. I hope it does.


uxmal:
Windows does indeed support '/' as a path separator. Thanks for the patch, and welcome back!

PR #26, opened 2015-08-21
Fix analysis unit tests - linux

nemerle:
Mostly /r/n issues + one filename case difference.

Seems one of the committed files had different line endings though :cry:

### DIFF
--- a/src/UnitTests/Analysis/IntraBlockDeadRegistersTests.cs
+++ b/src/UnitTests/Analysis/IntraBlockDeadRegistersTests.cs
@@ -36,9 +36,9 @@ namespace Reko.UnitTests.Analysis
     public class IntraBlockDeadRegistersTests
     {
         private string testResult;
-        private string toExpectedString(string [] lines)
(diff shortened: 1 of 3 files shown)


uxmal:
Thanks for this. Could you please adhere to the coding style of the code base, which is: all methods, properties, and events of classes and structs are **C**apitalized, regardless of whether they are public or private.

uxmal:
Also, consider use the `params` keyword to avoid having to create string arrays manually; ie. instead of
```c#
public void foo(string [] args) // ...

foo(new[] { "hello", "world" });
```
instead write
```c#
public void foo(params string [] args) // ...

foo("hello", "world");
```
You'll see examples of this sprinkled liberally in the reko codebase.

nemerle:
will do

nemerle:
done.. 

Also please note that SsaTransformTests and TrashedRegsiterFinder2Tests fail on linux. - this is related to #27, just converting their sources to unix line endings fixes them.

PR #40, opened 2015-08-24
C2xml extended usage text

nemerle:
Seems our ansi-c parser doesn't handler '#' lines at all so preprocessing has to be done in such a way as to remove those.


uxmal:
Indeed, this is a deliberate choice. Running a compiler preprocessor addresses this. I will take this change and tweak the usage message a little.

PR #41, opened 2015-08-24
IAT and ILT marked as data tables

uxmal:


halsten:
Please correct me if I understood this correctly, IAT and ILT will be always explicitly marked as data? If so, then this is not a good idea, as IAT can also hold code.

uxmal:
No fear.  The extent of the IAT is marked as (array (ptr code)) Actually , do you think making the ILT  (array ( opaque _ word 32)) is preferable?

halsten:
I think so, makes more sense to me this way.

PR #69, opened 2015-11-02
Remove Arch.x86 dependecy from IdentifierTests

nemerle:
Not sure if the reduction in tests 'strength' is acceptable, although I'm not sure this test was meant to also verify ```Registers.eax``` behavior ?


uxmal:
The eax instance here is not the "real" eax from Reko.Arch.X86, so the reduction is merited. Merging.

PR #88, opened 2015-12-18
TypeFormatter writes type reference without variable name

ptomin:
If variable type is pointer to type reference then TypeFormatter shoud write it as "{type reference name} * {variable name}". But it writes "{type reference name}" instead.

This request should fix it. New unit test was added.


uxmal:
Looks good. I especially appreciate the unit test.

PR #93, opened 2015-12-26
Fix SlibReadRealModeIntServices test

ptomin:
SlibReadRealModeIntServices was broken after creation msdos rename file service.
Number of services was increased from 88 to 89. So this test should be modified


uxmal:
Thanks again for keeping me honest.

Issue #99, opened 2016-01-09
UserSignatureBuilder does not resolve type references


ptomin:
e.g. if CSignature = "BYTE testFunc(BYTE a, BYTE b)" then UserSignatureBuilder throws CParserException exception. Type BYTE was not found but it was defined in windows32.xml.

I think, UserSignatureBuilder should pass type defintions were obtained from *.xml files to CParser
(Typedefs property of ParserState should be set)

uxmal:
Good catch. Reko should first load all type definitions, and then use them to resolve type references.

ptomin:
My suggestion to fix it : 8afa035

uxmal:
Next problem is that the `ProcedureSerializer` in use by `UserSignatureBuilder` is returning, on Win32,  `LPSTR` as a typedef of `<unknowntype>`, where it should be `typedef CHAR * LPSTR`. Working on it...

ptomin:
Yes, I known this problem. New constructor of TypeLibraryLoader should be added so that it can get predefined types

uxmal:
The `TypeLibraryLoader` class needs to be altered, indeed. It should be more like an "augmenter", which takes as input a `TypeLibrary`, which may or may not already contain types, and then sucks in the contents of the *serialized* type library and adds types and functions to the `TypeLibrary` that was passed in. Watch this space for more news....

uxmal:
OK, Pavel, please try the master branch after commit 56e0fac to see if this issue has finally been resolved.

ptomin:
1. UserSignatureBuilder works correctly with type references. Thanks!
2. Should unit test(s) be added for new functionality( Now CreateProcedureSerializer() returns procedure serialised populated with the types defined by the platform )?
3. What about user-defined (not system) types? Could user add them to project?

uxmal:
1. Glad to finally have knocked this out of the park!
2. If you identify the need for unit tests, don't hesistate to write them and submit PR's  <g>
3. There is a feedback loop here. A user runs reko on a binary and it discovers structures. Today, these are imginatively called `Eq_1`, `Eq_2` etc. At the end of the decompilation run, user should be able to edit the names and fields of the data structures, whereupon a restart of reko should pick up the new names. 

There should be a way for a user to add custom data types to the project, the question is how. Adding a C header file should be easy enough; reko has an ANSI C compiler to handle that. Is that enough? Does reko need a full blown source code editor, or something in between.

[...]

PR #105, opened 2016-01-13
Resolve type references

ptomin:
Now it is time to merge resolve-type-references with master
Thanks for contribution!


uxmal:
After a bit of an awkward merge, this is now checked in. Issue #99 is still open, however.

PR #106, opened 2016-01-15
Move common deserializer code to separate class

ptomin:
Move common deserializer code to separate class


uxmal:
I will make some tweaks as well. It is possible that unit tests using MockFactory will need to use other mock objects.

PR #111, opened 2016-01-16
Remove Xlib.xml and libc.so.xml from repository

ptomin:
I think Xlib.xml and libc.so.xml should not be stored in repository as they are generated automatically
So they should be removed from repository and added to gitignore. If you agree, merge this PR


uxmal:
At this point C2xml builds successfully so these files no longer need to be checked in.

PR #112, opened 2016-01-17
Replace BuildDataType method with TypeLibraryDeserializer

ptomin:
SerializedType::BuildDataType method seems to have the same purpose as TypeLibraryDeserializer.
I think it could be deleted and replaced with TypeLibraryDeserializer


ptomin:
...But one of the unit test failed after this changes. It should be fixed

ptomin:
All unit test have passed. I have modified SerializedProjectTests. Also I have moved mock platform code from UserSignatureBuilderTests and CodeViewerPaneTests to MockFactory

PR #125, opened 2016-01-29
Create python module sample

ptomin:
Create python module sample


uxmal:
Thanks for this. A reasonably sized binary that exhibits interesting behaviour causing Reko to fail. We can use the sample as a test case to improve the analysis and structure phases.

Issue #141, opened 2016-02-04
GlobalDataWorkItem and GlobalDataWriter should respect structure offsets.


uxmal:
Merge request #139 introduced the class `GlobalDataWorkItem`. When given a structure that contains internal padding, like
```C
struct Variant {
    unsigned short typeTag;     /* two byte size will force padding */
    float (*pfn)();
};
```
`GlobalDataWorkItem` will go off the rails because it assumes all fields in a `StructureType` are end-to-end with no gaps. This is not the case for structures that contain "small" fields that require padding to align subsequent fields properly.

I've added a disabled unit test `ScannerTests.Scanner_GlobalData_StructWithPadding` which fails. To get it to pass, `GlobalDataWorkItem` needs to be changed to respect `StructureField.Offset`.

@ptomin : could you please take a look at this? Once you have `GlobalDataWorkItem` working, `GlobalDataWriter` needs similar treatment.

ptomin:
John, thanks for test case. I'll try to fix it

ptomin:
I have modified GlobalDataWorkItem and GlobalDataWriter so that they use StructureField.Offset when reading structure fields. John, please test it and close issue if everything is correct

uxmal:
Looks good.

PR #142, opened 2016-02-04
Use StructureField.Offset when reading structure fields (#141)

ptomin:
Use StructureField.Offset when reading structure fields


uxmal:
Nicely done!

PR #156, opened 2016-03-02
Implement SetPositionAsFraction

ptomin:
Implement SetPositionAsFraction


ptomin:
Oh, you have already implemented SetPositionAsFraction

uxmal:
I managed to write a more complicated SetPositionAsFraction while you wrote this. Could you merge into my recent commit and put your and my code inside of a #if directive. That is, something like :
```C#
#if SIMPLE
   <ptomin code>
#else
   <uxmal code>
#endif
```
Then we can evaluate the result. I suspect your fix is the better one, the performance difference will be minuscule in either case.

ptomin:
Done

uxmal:
Thanks for this. We'll want to experiment with both versions, but obviously simpler code is better if nothing else differs.

PR #161, opened 2016-03-08
[WIP] AmigaOS library selection support

nemerle:
This is just a mostly GUI layer for now.
There are a few issues to resolve before this is complete:

* Unloading 'patterns'  - this would require that each 'applied' pattern records it's application location, so that when pattern is unloaded/changed all functions that contain given location are re-decompiled
* Splitting exec.funcs into exec_v33.funcs/exec_v34.funcs/ etc.  
* associating more information for applied functions i.e.
```AbortIO(ioRequest/a1)``` could be augmented by loading in file named exec_v33.rh, which defines exec library types and contains the following declaration ```[[reko::reg("D0")]] BYTE AbortIO([[reko::reg("A1")]] struct IORequest * iORequest);```

### DIFF
--- a/src/Environments/AmigaOS.Design/AmigaOSProperties.Designer.cs
+++ b/src/Environments/AmigaOS.Design/AmigaOSProperties.Designer.cs
@@ -31,7 +31,10 @@
             this.label1 = new System.Windows.Forms.Label();
             this.label2 = new System.Windows.Forms.Label();
             this.lblKickstart = new System.Windows.Forms.Label();
-            this.comboBox1 = new System.Windows.Forms.ComboBox();
+            this.cmbKickstartVersions = new System.Windows.Forms.ComboBox();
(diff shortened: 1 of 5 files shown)


nemerle:
Ouch, I hoped, to flesh this out a bit before merging into master :)

No worries though, it should have no influence on existing functionality

uxmal:
Hoops, got too eager there I guess! Not sure why github presented this as a Pull Request. No worries, just add a comment when you _do_ want it merged into master.

nemerle:
Well, I've marked it [WIP] in the PR, since there's no other way of doing this kind of thing on gtihub

I'd welcome any comments about my approach of using Interactor/Control thing, though.

uxmal:
Generally, any interaction logic, like the handling of focus loss / button click events and the like should be handled by the Interactor (Interactor is an old term for "Control" in the Model-View-Controller triad)

PR #162, opened 2016-03-12
Add new test case to Mcdm_MoveForward

ptomin:
I have added new test case to Mcdm_MoveForward. Now output is "00041000" for this test case. John, do you agree that correct output is "00042000"? Could you look at it?


uxmal:
Yes, it should be 0042000. Do you want me to fix this or are you working on it already.

ptomin:
It is your implementation of MoveToLine, so I want you to fix it

uxmal:
I hope to get done with this soon. Pulling the request now.

PR #165, opened 2016-03-19
Add unit tests for MixedCodeDataModel::GetLineSpans

ptomin:
I have added two unit tests to reproduce problems with MixedCodeDataModel::GetLineSpans method.

John, is it correct that MemoryArea size is larger than ImageSegment? It causes exception when viewing code in pySample.dll from 0x10001900 address (see Mcdm_GetLineSpans_MemoryAreaIsLargerThanSegment to reproduce)

### DIFF
--- a/src/UnitTests/Gui/Windows/Controls/MixedCodeDataModelTests.cs
+++ b/src/UnitTests/Gui/Windows/Controls/MixedCodeDataModelTests.cs
@@ -152,6 +152,57 @@ namespace Reko.UnitTests.Gui.Windows.Controls
         }
 
         [Test]
+        public void Mcdm_GetLineSpans_MemoryAreaIsLargerThanSegment()
+        {
+            var addrBase = Address.Ptr32(0x40000);
+
+            var memText = new MemoryArea(Address.Ptr32(0x41000), new byte[100]);
+            var memData = new MemoryArea(Address.Ptr32(0x42000), new byte[8]);
+            this.imageMap = new ImageMap(
+                addrBase,
+                new ImageSegment(".text", memText, AccessMode.ReadExecute) { Size = 4 },
+                new ImageSegment(".data", memData, AccessMode.ReadWriteExecute));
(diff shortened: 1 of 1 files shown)


uxmal:
A `MemoryArea` can be associated with one or more `ImageSegments`. In the case of a MS-DOS executable, where the memory used by the binary is always "compact", the MS-DOS image loader yields:
```
+----------------+  +---------------+  +---------------+
| ImageSegment 1 |  | ImageSegment2 |  | ImageSegment3 |
+----------------+  +---------------+  +---------------+
        v                   v                  v
+------------------------------------------------------+
| MemoryArea                                           |
+------------------------------------------------------+
```
ELF and PE image may be very sparse, with "gaps" of several hundred megabytes of address space. A monolithic chunk of MemoryArea would still be possible but it would be a tremendous waste of address space:
```
+----------------+  +---------------+  +---------------+
| ImageSegment 1 |  | ImageSegment2 |  | ImageSegment3 |
+----------------+  +---------------+  +---------------+
        v                   v                  v
+----------------+  +---------------+  +---------------+
| MemoryArea 1   |  | MemoryArea 2  |  | MemoryArea 3  |
+----------------+  +---------------+  +---------------+
```
Both of these scenarios must be handled in Reko code. In particular, MemoryAreas may exceed in address space the ImageSegments they are associated with.

I will take a look at this PR later today.

ptomin:
> In particular, MemoryAreas may exceed in address space the ImageSegments

... so Mcdm_GetLineSpans_MemoryAreaIsLargerThanSegment test is correct and GetLineSpans should be modified to process this case

uxmal:
Yes; I will be working on that later today -- unless you wish to take stab at it of course.

uxmal:
Thanks for the unit tests. The f43083d commit fixes a bunch of crashes, but now scrolling to the bottom is blowing up. I will look at that tomorrow.

Issue #167, opened 2016-03-21
Reko should respect the encoding of a binary and allow user to override it


uxmal:
Today, Reko is stupid and assumes all string text is encoded as ASCII. The only exception is the ZX81 environment where the Sinclair character set is done correctly. Reko should be able to:

- [ ] automatically detect the codepage if the binary header contains metadata that describes this
- [x] provide a mechanism to allow the user to override the current text encoding (perhaps with a menu like `View` >> `Text Encoding`)
- [ ] save this setting in the *.dcproj file so it can be retrieved later.

uxmal:
This is a lot better:

![cp1251cyrillic](https://cloud.githubusercontent.com/assets/10154465/13924095/4ad56c28-ef83-11e5-9ea0-52835c0e3b7c.png)

ptomin:
Very good!

PR #168, opened 2016-03-21
Create small x86 PE executable containing Russian text

ptomin:
Create small x86 PE executable containing Russian text. Is it enough, John?


uxmal:
Great. I will use this binary as a sample for testing issue #167. I shouldn't have to worry about whether the code is in UTF-8 or an older 8-bit codepage, as long as the operating system (in my case Windows 8) can provide a decoder for it.

PR #172, opened 2016-03-26
Create procedure or global variable declaration text box

ptomin:
Create procedure or global variable declaration text box:
sample1
Now it is possible to edit procedure declarations. Global variable declaration edit/creation has not implemented yet.

### DIFF
--- a/src/Gui/Windows/CombinedCodeViewInteractor.cs
+++ b/src/Gui/Windows/CombinedCodeViewInteractor.cs
@@ -156,7 +156,6 @@ namespace Reko.Gui.Windows
             this.combinedCodeView.MixedCodeDataView.VScrollValueChanged += MixedCodeDataView_VScrollValueChanged;
             this.combinedCodeView.MixedCodeDataView.Services = services;
             this.combinedCodeView.MixedCodeDataView.MouseDown += MixedCodeDataView_MouseDown;
-            this.combinedCodeView.MixedCodeDataView.MouseUp += MixedCodeDataView_MouseUp;
 
             this.combinedCodeView.CodeView.VScrollValueChanged += CodeView_VScrollValueChanged;
             this.combinedCodeView.CodeView.Services = services;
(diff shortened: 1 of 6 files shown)


src/Gui/Windows/CombinedCodeViewInteractor.cs
```diff
@@ -299,12 +304,39 @@ public bool ChooseTextEncoding()
             return true;
         }
 
-        private void MixedCodeDataView_MouseDown(object sender, EventArgs e)
+        private void MixedCodeDataView_MouseUp(object sender, MouseEventArgs e)
+        {
```
uxmal: I'd be happier with this change if it occurred after an explicit user action, rather than always on left-mouse-up. Consider defining a context menu for the MixedCodeDataView, and when the user selects the context menu, display the declarationTextBox. See `LowLevelnteractor.MarkType()` for an example. You can even define a shortcut key (`Ctrl+T` is used by the `LowLevelInteractor` to mark types.

In fact, as I write this, it makes total sense to integrate TypeMarker and DeclarationTextBox; they're almost doing the same thing, except that TypeMarker is using a dialect of "[Hungarian](https://en.wikipedia.org/wiki/Hungarian_notation)" for speed and convenience.

ptomin: I have added "Edit declaration" context menu item (see 25ecba6)

uxmal: Looks good; once you've addressed my comment below (regarding making the Ctrl+D accelerator specific to the CombinedCodeViewInteractor) I will merge the PR.

ptomin: > In fact, as I write this, it makes total sense to integrate TypeMarker and DeclarationTextBox; they're almost doing the same thing, except that TypeMarker is using a dialect of "Hungarian" for speed and convenience.

John, what do you mean? Now I'm going to implement global variable declaration using C Parser. Is it enough? Or you suggest to use Hungarian instead (or in addition to).

uxmal: I don't want to lose the ability to use hungarian to quickly denote the type of a range of memory. It's faster for me to  select a memory range and type `apsz` (array of pointers to zero-terminated strings) than `char * foo [124]`.

I suggest you continue working on the CombinedCodeView until you reach a stage where it is reasonably complete, and in the meantime I will give some thought to how these could be combined.

ptomin: OK, I see

PR #174, opened 2016-03-28
Add CParser test with incorrect statement caused infinite loop

ptomin:
Following CParser test causes infinite loop. John, could you look at it when you will have spare time?

### DIFF
--- a/src/UnitTests/Core/CLanguage/CParserTests.cs
+++ b/src/UnitTests/Core/CLanguage/CParserTests.cs
@@ -1136,5 +1136,18 @@ MemoryBarrier (
             var sExp = "(decl Char ((init-decl (ptr (func get ((Int n)))))))";
             Assert.AreEqual(sExp, decl.ToString());
         }
+
+        [Test]
+        public void CParser_IncorrectStatement()
+        {
+            Lex("typedef a b;");
+            Decl decl;
+            try {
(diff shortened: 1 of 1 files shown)


src/UnitTests/Core/CLanguage/CParserTests.cs
```diff
@@ -1136,5 +1136,18 @@ public void CParser_Regression()
...
+        {
+            Lex("typedef a b;");
+            Decl decl;
+            try {
+                decl = parser.Parse_ExternalDecl();
+            } catch (Exception) {
+                decl = null;
+            }
+            Assert.IsNull(decl);
+        }
```
uxmal: @ptomin: I have a fix for this, but I need to understand why you consider this an incorrect statement. The issue is  that in this fragment of C, the type `a` is not defined. However, had the fragment looked like this:
```C
typedef int a;
typedef a b;
```
it would be legal. 

I argue that the fragment `typedef a b;` should be legally parsed, with `a` becoming a type reference to an unknown type `a`. Later phases of Reko must always check that type references are properly resolved. 

Pavel, would you rather that the parser should signal an error since `a` hasn't been defined yet?

ptomin: I' m not sure that it should be illegal. I just want to fix for infinite loop. So you could fix it as you think to be correct.

PR #179, opened 2016-04-02
Parse global variable declarations

ptomin:
Modify DeclarationTextBox so that it allows to create/modify user global variables

### DIFF
--- a/src/Core/ImageMap.cs
+++ b/src/Core/ImageMap.cs
@@ -216,6 +216,37 @@ namespace Reko.Core
             item.Size = (uint)delta;
         }
 
+        public void RemoveItem(Address addr)
+        {
+            ImageMapItem item;
+            if (!TryFindItemExact(addr, out item))
+                return;
+
+            item.DataType = new UnknownType();
+
+            ImageMapItem mergedItem = item;
+
+            // Merge with previous item
+            ImageMapItem prevItem;
+            if (items.TryGetLowerBound((addr - 1), out prevItem) &&
+                prevItem.DataType is UnknownType)
+            {
+                mergedItem = prevItem;
+
+                mergedItem.Size = (uint)(item.EndAddress - mergedItem.Address);
+                items.Remove(item.Address);
+            }
+
+            // Merge with next item
+            ImageMapItem nextItem;
+            if (items.TryGetUpperBound((addr + 1), out nextItem) &&
+                nextItem.DataType is UnknownType)
+            {
+                mergedItem.Size = (uint)(nextItem.EndAddress - mergedItem.Address);
+                items.Remove(nextItem.Address);
+            }
+        }
+
         public ImageSegment AddSegment(MemoryArea mem, string segmentName, AccessMode mode)
         {
             var segment = new ImageSegment(
--- a/src/Core/Program.cs
+++ b/src/Core/Program.cs
@@ -120,6 +120,20 @@ namespace Reko.Core
             return up;
         }
 
+        public GlobalDataItem_v2 EnsureUserGlobal(Address address)
+        {
+            GlobalDataItem_v2 gbl;
+            if (!User.Globals.TryGetValue(address, out gbl))
+            {
+                gbl = new GlobalDataItem_v2()
+                {
+                    Address = address.ToString(),
+                };
+                User.Globals.Add(address, gbl);
+            }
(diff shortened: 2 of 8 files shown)


src/Core/ImageMap.cs
```diff
@@ -216,6 +216,37 @@ public void TerminateItem(Address addr)
...
+            // Merge with next item
+            ImageMapItem nextItem;
+            if (items.TryGetUpperBound((addr + 1), out nextItem) &&
+                nextItem.DataType is UnknownType)
+            {
+                mergedItem.Size = (uint)(nextItem.EndAddress - mergedItem.Address);
+                items.Remove(nextItem.Address);
+            }
+        }
+
```
uxmal: I'd like to see unit tests covering this new method on the central ImageMap class.

src/Core/Program.cs
```diff
@@ -120,6 +120,41 @@ public Serialization.Procedure_v1 EnsureUserProcedure(Address address, string na
...
+                DataType = dt,
+            };
+            if (size != 0)
+                this.ImageMap.AddItemWithSize(address, item);
+            else
+                this.ImageMap.AddItem(address, item);
+
+            return gbl;
+        }
+
```
uxmal: This method also needs some unit tests to capture the expected behaviour.

src/Decompiler/Analysis/UserSignatureBuilder.cs
```diff
@@ -160,5 +161,52 @@ public ProcedureBase_v1 ParseFunctionDeclaration(string fnDecl, Frame frame)
...
+                    Name = name,
+                    DataType = type,
+                };
+                return global;
+            }
+            catch (Exception ex)
+            {
+                //$TODO: if user has supplied a global variable that can't parse,
+                // we must notify them in the diagnostics window with a 
+                // WARNING.
```
uxmal: If this is happening due to a user action in the user interface, it's a good idea to indicate this immediately, by, say, turning the color red, flashing, or likewise.

src/Decompiler/Analysis/UserSignatureBuilder.cs
```diff
@@ -160,5 +161,52 @@ public ProcedureBase_v1 ParseFunctionDeclaration(string fnDecl, Frame frame)
...
+                Debug.Print("{0}\r\n{1}", ex.Message, ex.StackTrace);
+                return null;
+            }
+        }
+
+        public static bool IsValidCIdentifier(string id)
+        {
+            return Regex.IsMatch(id, "^[_a-zA-Z][_a-zA-Z0-9]*$");
+        }
```
uxmal: Could you add a //$REFACTOR comment here, like 

//$REFACTOR: this is language and platform dependent.

src/Decompiler/Analysis/UserSignatureBuilder.cs
ptomin: I have already done it in DeclarationTextBox::EnableControls. Is it not enough?

src/Decompiler/Analysis/UserSignatureBuilder.cs
ptomin: To IsValidCIdentifier? OK, it will be done

src/Decompiler/Analysis/UserSignatureBuilder.cs
uxmal: What I mean is that since you are changing the color as the user is typing, there is no need to litter the diagnostics with lots of warnings... Unless you generate a warning, like
```'foo@dbl' is not a valid C identifier```
and as the user types / corrects, you update the warning
```'foo@' is not a valid C identifier```
until the final @ is removed and the error is removed from the diagnostics list.

src/Core/ImageMap.cs
ptomin: Done in 3907b37

src/Decompiler/Analysis/UserSignatureBuilder.cs
ptomin: Done in  f2814ee

src/Decompiler/Analysis/UserSignatureBuilder.cs
ptomin: Done in 61137d0

PR #182, opened 2016-04-03
Process MapChanged event instead of direct refreshing of CombinedCode…

ptomin:
…View

### DIFF
--- a/src/Core/ImageMap.cs
+++ b/src/Core/ImageMap.cs
@@ -245,6 +245,8 @@ namespace Reko.Core
                 mergedItem.Size = (uint)(nextItem.EndAddress - mergedItem.Address);
                 items.Remove(nextItem.Address);
             }
+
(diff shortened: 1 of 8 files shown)


uxmal:
One general comment: any handler of the `ImageMap.MapChanged` that interacts with the user interface needs to call `Control.Invoke` or `Control.BeginInvoke`. The reason is that `ImageMap.MapChanged` may get raised in the background thread, which mustn't interact directly with the UI.

Of the methods `Invoke` and `BeginInvoke`, the better choice is `BeginInvoke` as generally, the `MapChanged` event is a fire-and-forget event. We don't want to tie up the decompiler engine while the .NET runtime synchronizes with the UI thread and it laboriously repaints itself.

ptomin:
See 222f28a. Is it correct now?

uxmal:
Looks good.

PR #185, opened 2016-04-05
CombinedCodeViewInteractor: add possibility to show only globals

ptomin:
CombinedCodeViewInteractor: add possibility to show only globals


src/Gui/Windows/CombinedCodeViewInteractor.cs
```diff
@@ -50,6 +50,9 @@ public class CombinedCodeViewInteractor : IWindowPane, ICommandTarget
 
         private DeclarationTextBox declarationTextBox;
 
+        private ImageSegment segment;
+        private bool hideProcedures;
```
uxmal: I prefer using "positive" property and field names whevener possible: prefer `showProcedures` to `hideProcedures` or `noShowProcedures`

src/Gui/Windows/CombinedCodeViewInteractor.cs
```diff
@@ -66,6 +69,7 @@ public void DisplayProcedure(Program program, Procedure proc)
         {
             this.program = program;
             this.proc = proc;
+            this.hideProcedures = false;
```
uxmal: `showProcedures = true;`
...etc

src/Gui/Windows/CombinedCodeViewInteractor.cs
ptomin: Fixed

PR #189, opened 2016-04-08
Add ProjectBrowserService test without global variables

ptomin:
Add ProjectBrowserService test without global variables


src/UnitTests/Gui/Windows/ProjectBrowserServiceTests.cs
```diff
@@ -393,12 +387,25 @@ private void Given_ProgramWithOneSegment()
             project.Programs.Add(program);
         }
 
+        private void Given_ImageMapItem()
+        {
```
uxmal: Consider refactoring this to `Given_ImageMapItem(uint address)` so this could be used in tests with multiple globals.

ptomin: Done in d542cc7

PR #191, opened 2016-04-10
Remove MixedCodeDataViewInteractor and MixedCodeDataViewService

ptomin:
John, are you going to use MixedCodeDataViewInteractor/MixedCodeDataViewService in future? If not it should be removed. CombinedCodeViewInteractor class is used instead now.


uxmal:
The service is obsolete so it should go.

PR #192, opened 2016-04-10
TextView: remove override of OnPaintBackground method

ptomin:
... to avoid drawing of black rectangle at the end
sample1

John, you have added overriding of OnPaintBackground in 6f3c6e8e. Mentioned reason is avoiding flicker. But there is no flicker after removing it. And there is no black rectangle at the end of TextView

### DIFF
--- a/src/Gui/Windows/Controls/TextView.cs
+++ b/src/Gui/Windows/Controls/TextView.cs
@@ -276,12 +276,6 @@ namespace Reko.Gui.Windows.Controls
             GetStyleStack().PopStyle();
         }
 
-        // Disable background painting to avoid the horrible flicker.
-        // We draw our own background anyway.
-        protected override void OnPaintBackground(PaintEventArgs pevent)
-        {
-        }
-
         public TextPointer GetStartSelection()
         {
             if (layout.ComparePositions(cursorPos, anchorPos) <= 0)


uxmal:
Pavel, you must have a faster computer than me! I definitely see a flicker when rendering. I'm working on painting background explicitly after the LineSpans if necessary.

BTW: it seems you have a non-proportional font rendering the MixedCodeView on the left -- is that intentional? My corresponding view looks lke this:
fixed-font-size
Reko is using the Lucida Console font here.

uxmal:
Commit 71548361 fixes the black background without introducing flicker.

ptomin:
> Pavel, you must have a faster computer than me! I definitely see a flicker when rendering. I'm working on painting background explicitly after the LineSpans if necessary.

OK, I'm waiting for you. So this PR should be closed

> BTW: it seems you have a non-proportional font rendering the MixedCodeView on the left -- is that intentional

No, it is not intentional. It is trouble for me to have such untidy view. I'll try to investigate it tomorrow

ptomin:
> Commit 7154836 fixes the black background without introducing flicker.
Thanks. Now it's OK

ptomin:
> Reko is using the Lucida Console font here.

it is using Micrososft Sans Serif on my machine. I have removed Reko node from windows registry but it is using Micrososft Sans Serif again.

uxmal:
Odd. Do you have Lucida Console installed on your machine?

uxmal:
You should be able to pick the font; what happens when you to to Tool / Options menu?

ptomin:
Yes, I have. I could set it using User Preferences dialog. But it resets back to  Micrososft Sans Serif after Reko restart

ptomin:
..and paddings between image items disappears after changing font to Lucida Console

ptomin:
John, maybe you need to remove Reko info from registry to reproduce this

PR #194, opened 2016-04-11
CombinedCodeViewInteractor::copy: do not drop the last character

ptomin:
It does not drop the last character after this commit. I do not know why.


ptomin:
There other problems with copy/paste. 
1. Exception is thrown if start selection position is at the end of line
2. If end  selection position is at the end of line then it copies all data from start selection to the end of TextView
I'll try to fix them later

uxmal:
sounds good.

PR #201, opened 2016-04-17
Add unit test for MixedCodeDataModel

ptomin:
This unit test failed. It seems to be bug in GetLineSpans method. John, could you look into it?


uxmal:
I'll peek at it.

uxmal:
Let me know if that fixes the issue. GetLineSpans is getting crufty and may benefit from a rewrite, as long as it passes all the tests.

ptomin:
It works now. Thanks!

PR #215, opened 2016-04-24
Add _assert signature to msvcrt.xml

ptomin:
I have added assert function signature 
`void _assert(char * __cond, char * __file, int __line)`
to msvcrt.xml
John, is it right place for this function?


uxmal:
Yes it is correct.

PR #216, opened 2016-04-29
TypeTransformer: do not transform user-defined structs and unions

ptomin:
I have modified DataTypeTransformer so that it does not transform user-defined structs and unions in a421e30. John, should be the same changes applied to TypeTransformer?

### DIFF
--- a/src/Decompiler/Typing/TypeTransformer.cs
+++ b/src/Decompiler/Typing/TypeTransformer.cs
@@ -60,7 +60,8 @@ namespace Reko.Typing
 			this.eventListener = eventListener;
 			this.unifier = new Unifier(factory);
 			this.comparer = new DataTypeComparer();
-		}
+            this.visitedTypes = new HashSet<DataType>();
+        }
 
 		public bool Changed
 		{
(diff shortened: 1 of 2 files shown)


uxmal:
Indeed the type transformer should leave user defined types alone. Could you please generate a unit test to verify that this indeed is happening?

ptomin:
> Could you please generate a unit test to verify that this indeed is happening?

Done in 6e148a4

uxmal:
Pavel, this wasn't quite what I was looking for. Rather than making the Fields and Alternatives properties virtual, try the following: a structure with a single field at offset 0 of type `T1` will be simplified by the type transformer to just `T1`. However, if the user has created a user-defined structure
```C
struct T2 {
     struct T1 field0000;
}
```
then this should remain unaffected. So the unit test should be something like:
```C#
[Test]
public void TtranUserStruct() {
    var t1 = new StructureType("T1", 4, true); 
    var t2 = new StructureType("T2", 0, true);
   var ttran = new TypeTransformer(....);
   var dt = t2.Accept(ttran);
   Assert.AreSame(t2, dt, "Should not affect user-defined types");
}

[Test]
public void TtranNonUserStruct() {
    var t1 = new StructureType("T1", 4, true); 
    var t2 = new StructureType("T2", 0, true);
   var ttran = new TypeTransformer(....);
   var dt = t2.Accept(ttran);
   Assert.AreSame(t1, dt, "Should reduce fields at offset 0 ");
}
```

ptomin:
I have added TtranUserStruct and TtranNonUserStruct in 9a7b293

PR #217, opened 2016-04-30
Do not scan procedures marked as no decompile

ptomin:
Scanner should not scan procedures marked as no decompile

### DIFF
--- a/src/UnitTests/Scanning/ScannerTests.cs
+++ b/src/UnitTests/Scanning/ScannerTests.cs
@@ -518,6 +518,79 @@ fn00001100_exit:
         }
 
         [Test]
+        public void Scanner_NoDecompledProcedure()
+        {
+            Given_Program(Address.Ptr32(0x1000), new byte[0x2000]);
+            program.User.Procedures.Add(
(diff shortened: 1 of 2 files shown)


src/UnitTests/Scanning/ScannerTests.cs
```diff
@@ -518,6 +518,79 @@ public void Scanner_Trampoline()
...
+                    Decompile = false,
+                }
+            );
+
+            var sc = CreateScanner(program);
+            sc.EnqueueEntryPoint(
+                new EntryPoint(
+                    Address.Ptr32(0x12314),
+                    program.Architecture.CreateProcessorState()));
+            sc.EnqueueEntryPoint(
```
uxmal: Are the two EntryPoints intentional? If so, add a comment explaining why there are two entry points with same address, otherwise remove the duplicate.

ptomin: There are two different entry points: 0x12314 and 0x12324

uxmal: OK, I misread. Looks good.

PR #219, opened 2016-05-01
Enable StructWithPadding + Replace array length assertion + Add null check for procedure

ptomin:
3 little fixes:
 - GlobalDataWorkItemTests: enable StructWithPadding test
 - GlobalDataWriter: replace array length assertion with decompiler warning
 - Cmd_EditSignature: add null pointer check for procedure


uxmal:
looks good

PR #220, opened 2016-05-01
Implement ApplyConstants for LeOperator and GeOperator

ptomin:
- Implement ApplyConstants for LeOperator and GeOperator
- Add unit tests to verify it

### DIFF
--- a/src/UnitTests/Core/OperatorTests.cs
+++ b/src/UnitTests/Core/OperatorTests.cs
@@ -198,9 +198,81 @@ namespace Reko.UnitTests.Core
 			Assert.IsNotNull(new AddressOfOperator());
 		}
 
+        [Test]
+        public void LtApplyConstants1()
+        {
+            RunApply(5, 3, new LtOperator(), 0);
+        }
+
+        [Test]
+        public void LtApplyConstants2()
+        {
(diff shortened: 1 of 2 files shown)


src/UnitTests/Core/OperatorTests.cs
```diff
@@ -198,9 +198,81 @@ public void AddressOfCreate()
...
+            RunApply(4, 5, new GeOperator(), 0);
+        }
+
+        [Test]
+        public void GeApplyConstants3()
+        {
+            RunApply(6, 6, new GeOperator(), 1);
+        }
+
+
```
uxmal: Need unit tests to handle Constant.Invalid as well

ptomin: What do you mean?

uxmal: Never mind, I'll add those after I take the PR

ptomin: .. you implementation of LtOperator and GtOperator does not handle Constant.Invalid. Is it correct?

uxmal:
They all need to handle it. I'm adding support for it right now.

PR #221, opened 2016-05-01
ImageMap::RemoveItem: do not merge disjoined items

ptomin:
- ImageMap::RemoveItem: do not merge disjoined items
- Add unit test to verify it

### DIFF
--- a/src/Core/ImageMap.cs
+++ b/src/Core/ImageMap.cs
@@ -230,7 +230,8 @@ namespace Reko.Core
             // Merge with previous item
             ImageMapItem prevItem;
             if (items.TryGetLowerBound((addr - 1), out prevItem) &&
-                prevItem.DataType is UnknownType)
+                prevItem.DataType is UnknownType &&
+                prevItem.EndAddress.Equals(item.Address))
             {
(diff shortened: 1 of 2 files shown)


src/Core/ImageMap.cs
```diff
@@ -241,7 +242,8 @@ public void RemoveItem(Address addr)
             // Merge with next item
             ImageMapItem nextItem;
             if (items.TryGetUpperBound((addr + 1), out nextItem) &&
-                nextItem.DataType is UnknownType)
+                nextItem.DataType is UnknownType &&
+                mergedItem.EndAddress.Equals(nextItem.Address))
             {
```
uxmal: The "==" operator is overloaded for `Address`, so you should feel OK to use it..

ptomin: I have not seen "==" implementation. And some tests are failed when I try to use "==".

uxmal:
I'm having an aneurysm here: I could have sworn it was implemented a _long_ time ago; all the other comparison operators are implemented after all. I'll take the PR and then investigate what happened to the '==' and '!=' operators.

PR #222, opened 2016-05-02
Add unit tests for IdConstant to catch problem with type references

ptomin:
IdConstant fails to transform identifiers with reference type. I have created unit tests to catch it. John, could you look at it?


uxmal:
Try this now.

ptomin:
> Try this now.

Now it works. Thanks!

PR #232, opened 2016-05-08
Unifier: keep name of structure field

ptomin:
Now `Unifier` could lose name of global variable specified by user.
- Keep name of structure field
- Add unit test to verify it

### DIFF
--- a/src/Core/Types/StructureField.cs
+++ b/src/Core/Types/StructureField.cs
@@ -51,7 +51,7 @@ namespace Reko.Core.Types
         public override string Name { get { if (name == null) return GenerateDefaultName(); return name; }
             set { name = value; } }
 
-        public bool NameSet { get { return name != null; } }
+        public bool IsNameSet { get { return name != null; } }
 
         private string name;
 
--- a/src/Core/Types/StructureField.cs
+++ b/src/Core/Types/StructureField.cs
@@ -50,6 +50,9 @@ namespace Reko.Core.Types
 
         public override string Name { get { if (name == null) return GenerateDefaultName(); return name; }
             set { name = value; } }
+
+        public bool NameSet { get { return name != null; } }
+
         private string name;
 
         public int Offset { get; set; }
--- a/src/Core/Types/Unifier.cs
+++ b/src/Core/Types/Unifier.cs
@@ -506,8 +506,9 @@ namespace Reko.Core.Types
 				}
 				else
 				{
-                    DataType fieldType = Unify(fa.DataType, fb.DataType);
-					mem.Fields.Add(fa.Offset, fieldType);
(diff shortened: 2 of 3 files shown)


src/Core/Types/StructureField.cs
```diff
@@ -50,6 +50,9 @@ public StructureField(int offset, DataType type, string name)
 
         public override string Name { get { if (name == null) return GenerateDefaultName(); return name; }
             set { name = value; } }
+
+        public bool NameSet { get { return name != null; } }
+
```
uxmal: Please rename the property to `IsNameSet`. I like boolean predicates to indicate their predicateness, and sadly C#, unlike scheme, doesn't allow `NameSet?` as a property name.

src/Core/Types/Unifier.cs
```diff
@@ -534,6 +535,17 @@ public DataType UnifyStructures(StructureType a, StructureType b)
 			return mem;
 		}
 
+        private string MakeFieldName(StructureField fa, StructureField fb)
+        {
+            if (fa.NameSet && fb.NameSet && fa.Name != fb.Name)
+                throw new NotSupportedException("Both structure fields have names");
+            if (fa.NameSet)
```
uxmal: It would be nice if there were more context in the exception so that trouble shooting this is easier. For instance, change this to 'bool TryMakeFieldName(fa, fb, out string name)`. Have it return `false` if it fails because of name conflict. Then make the caller of `TryMakefieldName` throw an exception to the effect of "Failed to unify field 'xxxx' in stucture 'yyyy' with field 'zzzz' in structure 'www'."

src/Core/Types/StructureField.cs
ptomin: Done in 05cbe8f

src/Core/Types/Unifier.cs
ptomin: Done in 881de40

src/Core/Types/Unifier.cs
```diff
@@ -507,7 +507,12 @@ public DataType UnifyStructures(StructureType a, StructureType b)
 				else
 				{
                     var fieldType = Unify(fa.DataType, fb.DataType);
-                    var fieldName = MakeFieldName(fa, fb);
+                    string fieldName;
+                    if (!TryMakeFieldName(fa, fb, out fieldName))
+                        throw new NotSupportedException(
+                            string.Format(
+                                "Failed to unify field '{0}' in structure '{1}' with field '{2}' in structure '{3}'.",
+                                fa.Name, a, fb.Name, b));
```
uxmal: Pleas add a unit test that verifies that this exception is getting thrown correctly.

ptomin: Done in c345e69

PR #249, opened 2016-06-03
Fix for expression simplifier type ref

ptomin:
- Apply the same fix as 1a0e1e2 and 7dcce04 to  ExpressionSimplifier.VisitCast
- Add unit test to verify it

### DIFF
--- a/src/Decompiler/Evaluation/ExpressionSimplifier.cs
+++ b/src/Decompiler/Evaluation/ExpressionSimplifier.cs
@@ -329,7 +329,11 @@ namespace Reko.Evaluation
             if (exp == Constant.Invalid)
                 return exp;
 
-            var ptCast = cast.DataType as PrimitiveType;
+            var dtCast = cast.DataType;
+            var typeref = dtCast as TypeReference;
+            if (typeref != null)
+                dtCast = typeref.Referent;
+            var ptCast = dtCast as PrimitiveType;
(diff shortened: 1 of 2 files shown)


src/Decompiler/Evaluation/ExpressionSimplifier.cs
```diff
@@ -329,7 +329,11 @@ public virtual Expression VisitCast(Cast cast)
             if (exp == Constant.Invalid)
                 return exp;
 
-            var ptCast = cast.DataType as PrimitiveType;
+            var dtCast = cast.DataType;
+            var typeref = dtCast as TypeReference;
+            if (typeref != null)
+                dtCast = typeref.Referent;
+            var ptCast = dtCast as PrimitiveType;
```
uxmal: Consider rewriting this as 
```
ptCast = cast.DataType.ResolveAs<PrimitiveType>();
```

ptomin: What do you mean? Rewrite last line or replace all lines with `ResolveAs<PrimitiveType>`?

uxmal: I'm sorry, I'm confusing `TypeReference` with `TypeVariable` -- again. I need to come up with better names for those classes or find a way to unify them. Pulling now....

PR #253, opened 2016-06-13
Reintroduce function type deserialization

ptomin:
- Reintroduce function type deserialization
- Reenable Tlldr_signature unit test

### DIFF
--- a/src/Core/Types/FunctionType.cs
+++ b/src/Core/Types/FunctionType.cs
@@ -69,9 +69,34 @@ namespace Reko.Core.Types
             }
         }
 
-        public FunctionType(SerializedSignature signature) : base()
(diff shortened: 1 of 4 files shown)


src/Core/Types/FunctionType.cs
```diff
@@ -69,9 +69,34 @@ public FunctionType(ProcedureSignature sig) : base()
             }
         }
 
-        public FunctionType(SerializedSignature signature) : base()
+        static public FunctionType Create(ISerializedTypeVisitor<DataType> tldser, SerializedSignature signature)
```
uxmal: It seems like this method should be moved into the `TypeLibraryDeserializer.VisitSignature` class, since most of the work involves using the `tldser` variable.

ptomin: Done in 642b68c

PR #257, opened 2016-06-25
VCExeSample: create function with indirect call

ptomin:
VCExeSample: create function with indirect call


ptomin:
Also see https://github.com/uxmal/reko/commit/d4dc1aaa2a889e3a550ea64d6bc11fa178e52a73#commitcomment-18013103

uxmal:
Have you tried distilling the error into a unit test? Regardless  I will take a look this evening.

ptomin:
Maybe 04f4604 could be useful

ptomin:
.. see #258

ptomin:
Maybe b8ce737 could be possible fix for it

ptomin:
@uxmal what do you think about  b8ce737? It's not final version and should be improved, off course

PR #260, opened 2016-07-08
VCExeSample: add indirect call of global struct field

ptomin:


ptomin:
It could be fixed by 376c02b

PR #261, opened 2016-07-09
Global data inference

ptomin:
-  Modify `ExpressionTypeAscender `to do type inference with user global data
-  Create unit test to verify it


uxmal:
Very nice work, Pavel!

Issue #254, opened 2016-06-13
PE: Section names don't have to be unique, fails with dictionary mapping


Lukas-Dresel:
Section names in PE files are afaik not forced to be unique. I have an executable that has 11 sections of the same name. However, the PE loader uses a SortedDictionary which then invalidly seems to contain only 1 section instead of all 11 since the new section assignments overwrite the previous ones.

uxmal:
Great catch. It turns out that section names are merely "suggestions", and bad guys can set them all to the same name if they wish. Unfortunately, reko may be making a lot of assumtions on section name uniqueness. Will look at this tomorrow.

Lukas-Dresel:
Yeah, that's what the obfuscation scheme on my binary does. On the other hand we can maybe use this as a heuristic for the specific packer of that binary.

Lukas-Dresel:
I'll see if i can figure out what the obfuscation schemes are and if they are automatically "unpackable".

uxmal:
As far as I can gather, loader should avoid making decisions based on section names, as they are purely for human consumption (http://stackoverflow.com/questions/14418153/pe-file-section-names)  However, reko uses them in its user interface and so needs a mechanism to differentiate between sections that have identical names.

Lukas-Dresel:
If you want i can try take a look at it to get started. Since i have the binary here maybe i'll run into other stuff too

Lukas-Dresel:
If you want i can try take a look at it to get started. Since i have the binary here maybe i'll run into other stuff too

uxmal:
Be bold and forge ahead!

Some suggestions: have `PeImageLoader.LoadSections` ensure unique names by checking each loaded section name against a *private* dictionary, and if the name already exists, generate a new one, like "foo_1", "foo_2" etc. 

Add a property `OriginalName` to the internal `PeImageLoader.Section` class and the public `Reko.Core.ImageSegment` class. If `OriginalName` != `Name` the `Reko.Gui.Design.ImageMapSegmentNodeDesigner` class should set `TreeNode.Text` to `Foo_1 (Foo)` (that is, `string.Format("{0} ({1})", segment.Name, segment.OriginalName)` ) rather than simply `segment.Text`.

Those are the major points that come to mind at midnight (!). Let me know how I can help you get going.

Lukas-Dresel:
[...]

PR #262, opened 2016-07-13
Fix for issue #254

Lukas-Dresel:
Rewrote the sectionMap to a sectionList instead, 

Is there any place to put regression tests for cases like this? Or do you put normal UnitTests instead?


uxmal:
If you have a binary that exercises the issue you've fixed, and it is not encumbered by copyrigghts, licenses etc, it'd be great if you could place it under /subjects/regressions/reko-<issue / PR #>. Otherwise, you can create a unit test for it (in PeImageLoaderTests.cs) where you create an in-memory PE image with sections with the same names.

PR #264, opened 2016-07-13
X86 disass bugfix invalid state

Lukas-Dresel:
Rewrote the sectionMap to a sectionList instead,

Is there any place to put regression tests for cases like this? Or do you put normal UnitTests instead?

### DIFF
--- a/src/Arch/X86/X86Disassembler.cs
+++ b/src/Arch/X86/X86Disassembler.cs
@@ -34,21 +34,277 @@ namespace Reko.Arch.X86
 	/// </summary>
 	public partial class X86Disassembler : DisassemblerBase<X86Instruction>
 	{
+        private class X86LegacyCodeRegisterExtension
+        {
+            const byte MAGIC = 0x40;
+            const byte MAGIC_MASK = 0xf0;
+            internal static X86LegacyCodeRegisterExtension Disabled = new X86LegacyCodeRegisterExtension(0);
+            
+            byte val;
+
+            internal X86LegacyCodeRegisterExtension(byte value)
+            {
+                this.val = value;
+            }
+            internal X86LegacyCodeRegisterExtension(byte magic, bool wide, bool modrm_reg, bool sib_idx, bool modrm_rm)
+            {
+                this.val = (byte)((this.val & 0xf) | ((magic & 0xf) << 4));
+                this.WideValue = wide;
+                this.Target_ModRM_Reg = modrm_reg;
+                this.Target_SIB_Index = sib_idx;
+                this.Target_ModRM_RegOrMem = modrm_rm;
+            }
+
+            internal byte ByteValue
+            {
+                get
+                {
+                    return this.val;
+                }
+                set
+                {
+                    this.val = value;
+                }
+            }
+            internal bool Active
+            {
+                get
+                {
+                    return this.val != 0;
+                }
+            }
+            internal bool WideValue
+            {
+                get
+                {
+                    return ((this.val & 0x8) == 0x8);
+                }
+                set
+                {
+                    if (value)
+                    {
+                        this.val = (byte)(this.val | 0x8);
+                    }
+                    else
+                    {
+                        this.val = (byte)(this.val & ~0x8);
+                    }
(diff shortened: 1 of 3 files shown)


src/Arch/X86/X86Disassembler.cs
```diff
@@ -34,21 +34,277 @@ namespace Reko.Arch.X86
...
+            {
+                this.val = (byte)((this.val & 0xf) | ((magic & 0xf) << 4));
+                this.WideValue = wide;
+                this.Target_ModRM_Reg = modrm_reg;
+                this.Target_SIB_Index = sib_idx;
+                this.Target_ModRM_RegOrMem = modrm_rm;
+            }
+
+            internal byte ByteValue
+            {
```
uxmal: For compactness I prefer that properties that don't raise events be automatic properties, ie
internal byte ByteValue {get; set; }

src/Arch/X86/X86Disassembler.cs
```diff
@@ -34,21 +34,277 @@ namespace Reko.Arch.X86
...
+                {
+                    return this.val;
+                }
+                set
+                {
+                    this.val = value;
+                }
+            }
+            internal bool Active
+            {
```
uxmal: My preference is for boolean properties to have (where grammatically possible) an `Is`... prefix to make it clearer. So, please change this to `IsActive`.

src/Arch/X86/X86Disassembler.cs
```diff
@@ -34,21 +34,277 @@ namespace Reko.Arch.X86
...
+                get
+                {
+                    return this.val != 0;
+                }
+            }
+            internal bool WideValue
+            {
```
uxmal: Same idea here, `IsWideValue`

src/Arch/X86/X86Disassembler.cs
```diff
@@ -34,21 +34,277 @@ namespace Reko.Arch.X86
...
+                        this.val = (byte)(this.val | 0x8);
+                    }
+                    else
+                    {
+                        this.val = (byte)(this.val & ~0x8);
+                    }
+                }
+            }
+            internal bool Target_ModRM_Reg
+            {
```
uxmal: Reko's coding style has evolved over time (and I've never formally written it down), but I prefer no underscores in names except in the unit tests. I'd prefer `TargetModrmRegister`

src/Arch/X86/X86Disassembler.cs
```diff
@@ -34,21 +34,277 @@ namespace Reko.Arch.X86
...
+                        this.val = (byte)(this.val | 0x1);
+                    }
+                    else
+                    {
+                        this.val = (byte)(this.val & ~0x1);
+                    }
+                }
+            }
+            internal bool Target_SIB_Index
+            {
```
uxmal: IsTargetSibInddex

uxmal:
Lukas, for disassembler changes it is very important that you add unit tests that exercise your new functionality. Look at the class `X86DisassemblerTests` which contains unit tests for the X86 disassembler.

src/Arch/X86/X86Disassembler.cs
Lukas-Dresel: I'll change it to a method then, should have done that anyway, I don't like having Properties called Is...

src/Gui/Windows/Controls/TextViewModel.cs
```diff
@@ -99,10 +99,34 @@ public interface TextViewModel
...
-        public object CurrentPosition { get { return this; } }
-        public object StartPosition { get { return this; } }
-        public object EndPosition{ get { return this; } }
-        public int LineCount { get { return 0; } }
+        public object CurrentPosition
+        {
+            get
+            {
+                return null;
+            }
```
uxmal: Is there a reason why these were changed to `null`? I'd like to avoid injecting invalid objects into the live object graph where possible.

Lukas-Dresel:
Closed, relics of other temporary fixes contained, will open new one with a single commit.

PR #267, opened 2016-07-13
Write file name to header comment instead of full path

ptomin:
Now there is several differences after regression tests run on my machine. The reason is difference between full path on my and you machines. I suggest to write only file name to header so it will be easier  to compare output for regression tests.


uxmal:
Makes sense to not destabilize the regression tests. Great initiative, Pavel!

Issue #266, opened 2016-07-13
InvalidCastException in OnScroll in UI controls


Lukas-Dresel:
An InvalidCastException occurs when dragging in the hex views string area.

Reproduce:
1. Start reko
2. Open ~/subjects/PE-x86/VCExeSample/VCExeSample.exe
3. Select some bytes in the hexdump so that they are highlighted in blue (definitely works when selecting multiple lines, sometimes doesn't trigger when only selecting bytes in the first line)
4. Click and drag around in the string representation part of the hexdump
5. Exception happens in Gui/Windows/Controls/DisassemblyControl.cs line 119 OnScroll Method when the Model.CurPosition is cast to Address

Lukas-Dresel:
Similar issues could maybe arise in 

Gui/Windows/Controls/MixedCodeDataControl.cs:
- (74, 46) : addrTop = (Address)Model.CurrentPosition;
- (99, 38) : addrTop = (Address)Model.CurrentPosition;
- (105, 49) : var currentAddress = (Address)Model.CurrentPosition;
- (108, 43) : this.addrTop = (Address)model.CurrentPosition;

Line member (because it is assigned CurrentPosition at Gui\Windows\Controls\TextView.cs line 410)
- (131, 43) : return (Address)anchorPos.Line;

Gui\Windows\Controls\NestedTextModel.cs 
- (86, 38) : Location loc = (Location)CurrentPosition;

Gui\Windows\Controls\DisassemblyControl.cs 
- (119, 41) : topAddress = (Address)Model.CurrentPosition;

Lukas-Dresel:
Btw, the problem is that EmptyEditorModel.CurrentPosition cannot be cast to Address which it tries to do because:

Gui\Windows\Controls\TextViewModel.cs - (102,23) :
```
        public object CurrentPosition { get { return this; } }
        public object StartPosition { get { return this; } }
        public object EndPosition{ get { return this; } }
```

uxmal:
The issue is in MemoryControl.  I hope to have a fix tomorrow.

Lukas-Dresel:
For reference, see pull request and John's explanation at #269

uxmal:
Haven't heard from anyone about this for a while, so closing for now. If problem persists, please do reopen the issue.

PR #269, opened 2016-07-13
Possible fix for #266

ptomin:


Lukas-Dresel:
I don't think this fix is correct. What about the other instances of casting to Address? Wouldn't that still crash because Address and EmptyEditorModel are incompatible?

ptomin:
@Lukas-Dresel Workflow described in #266 does not cause exception after this fix. Are there other workflows  that cause cast exceptions?

@uxmal What do you think about it?

Lukas-Dresel:
Ah, i misunderstood the as-operator. What about the other listed instances?

uxmal:
Although this may fix the immediate issue  (crash) it doesn't fix the cause of the issue,  which is that when the user clicks on the alphanumeric dump, the memory control erroneously fires a selection event with a null address.  The "real " fix is to either not raise the event,  or better yet, fire the event with a valid address since the alphanumeric characters can all be mapped to addresses.

@ptomin : please check the other places where the position is being cast to address. Once you're satisfied update this PR. This will fix the symptom. However leave #266 open and I will fix the root issue.

@

ptomin:
@uxmal @Lukas-Dresel I have replaced casts with `as` operators. Now it's time to merge it

PR #278, opened 2016-07-21
Modify unifier

ptomin:
- Try to unify type variable rather than make union

e.g if `T_31 {dtOrig=word32 ,dt=word32}` will be unify with `int32` should be `T_31 {dtOrig=int32, dt=int32}`  rather than `union {T_31 {dtOrig=word32, dt=word32}, int32}`
- Try to unify pointer rather than make union

e.g if `T_32 {dtOrig=word32 ,dt=word32}` will be unify with `ptr(real32)` should be `T_32 {dtOrig=ptr(real32), dt=ptr(real32)}`  rather than `union {T_32 {dtOrig=word32, dt=word32}, ptr(real32)}`
- Unify void type
- Fix broken unit tests
- Run regression tests

Many output files for regression tests was changed. John, please review, is it progress or regression.


uxmal:
I will look at this once I have access to a computer again. Hopefully tomorrow!

uxmal:
Pavel, this looks very good. You've found a case in the unifier that I'd missed back in the early Neolithic that takes care of many spurious and unnecessary unions. Great job!

ptomin:
@uxmal What about unification between `structure {0 T_3 { dtOrig=word32 ,dt=word32 } } ` and `int32`. Shoud not it be `structure {0 T_3 { dtOrig=int32, dt=int32 } } ` ?

ptomin:
... e.g. 0417a82 does such unification. But it fails on PySample

PR #280, opened 2016-07-27
Indirect call rewriter

ptomin:
Create class which should try to rewrite indirect call statements to
Applications using user-defined data (e.g. global variables, parameters
of procedures).
It should be executed after ssa transform.
It used ExpressionTypeAscender to define type of callee.
If type of callee is function type it replaces call with application.
It also replaces parameters and return value with ssa identifiers

There is also task to include it to new ssa(analysis-development branch)

### DIFF
--- a/src/Decompiler/Analysis/IndirectCallRewriter.cs
+++ b/src/Decompiler/Analysis/IndirectCallRewriter.cs
@@ -34,7 +34,7 @@ namespace Reko.Analysis
     /// Try to rewrite indirect call statements to Applications using
     /// user-defined data (e.g. global variables, parameters of procedures).
     /// </summary>
-    class IndirectCallRewriter : ExpressionTypeAscender
+    public class IndirectCallRewriter : ExpressionTypeAscender
     {
         private SsaState ssa;
         private Program program;
--- a/src/Decompiler/Analysis/IndirectCallRewriter.cs
+++ b/src/Decompiler/Analysis/IndirectCallRewriter.cs
@@ -19,14 +19,12 @@
 #endregion
 
 using System;
-using System.Diagnostics;
 using System.Linq;
 using Reko.Core;
 using Reko.Core.Code;
 using Reko.Core.Expressions;
 using Reko.Core.Services;
 using Reko.Core.Types;
-using Reko.Typing;
 
 namespace Reko.Analysis
 {
@@ -34,7 +32,7 @@ namespace Reko.Analysis
     /// Try to rewrite indirect call statements to Applications using
     /// user-defined data (e.g. global variables, parameters of procedures).
     /// </summary>
-    public class IndirectCallRewriter : ExpressionTypeAscender
+    public class IndirectCallRewriter : ExpressionTypeAscenderBase
     {
         private SsaState ssa;
         private Program program;
@@ -46,7 +44,7 @@ namespace Reko.Analysis
             Program program,
             SsaState ssa,
             DecompilerEventListener eventListener) :
-            base(program, new TypeStore(), new TypeFactory())
+            base(program, new TypeFactory())
         {
             this.program = program;
             this.proc = ssa.Procedure;
@@ -105,6 +103,16 @@ namespace Reko.Analysis
                     return sid.DefExpression.Accept(this);
             }
             return base.VisitIdentifier(id);
+        }
+
+        protected override DataType RecordDataType(DataType dt, Expression exp)
+        {
+            return dt;
+        }
+
+        protected override DataType EnsureDataType(DataType dt, Expression exp)
+        {
+            return dt;
         }
     }
 
--- a/src/Decompiler/Analysis/IndirectCallRewriter.cs
+++ b/src/Decompiler/Analysis/IndirectCallRewriter.cs
@@ -1,0 +1,213 @@
+﻿#region License
+/* 
+ * Copyright (C) 1999-2016 Pavel Tomin.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ */
+#endregion
+
+using System;
+using System.Diagnostics;
+using System.Linq;
+using Reko.Core;
+using Reko.Core.Code;
+using Reko.Core.Expressions;
+using Reko.Core.Types;
+using Reko.Typing;
+
+namespace Reko.Analysis
+{
+    /// <summary>
+    /// Try to rewrite indirect call statements to Applications using
+    /// user-defined data (e.g. global variables, parameters of procedures).
+    /// </summary>
+    class IndirectCallRewriter : ExpressionTypeAscender
+    {
+        private SsaState ssa;
+        private Program program;
+        private Procedure proc;
+        private SsaIdentifierTransformer ssaIdTransformer;
+
+        public IndirectCallRewriter(Program program, SsaState ssa) :
+            base(program, new TypeStore(), new TypeFactory())
+        {
+            this.program = program;
+            this.proc = ssa.Procedure;
+            this.ssa = ssa;
+            this.ssaIdTransformer = new SsaIdentifierTransformer(ssa);
+        }
+
+        public void Rewrite()
+        {
+            foreach (Statement stm in proc.Statements)
+            {
+                CallInstruction ci = stm.Instruction as CallInstruction;
+                if (ci != null)
+                {
+                    try
+                    {
+                        RewriteCall(stm, ci);
+                    }
+                    catch (Exception ex)
+                    {
+                        Debug.Print(
+                            "Exception {0}" + Environment.NewLine +
+                            "when rewriting indirect call statement: {1}",
+                            ex, stm);
+                    }
+                }
+            }
+        }
+
+        private void RewriteCall(Statement stm, CallInstruction call)
+        {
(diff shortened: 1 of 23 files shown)


src/Decompiler/Analysis/IndirectCallRewriter.cs
```diff
@@ -0,0 +1,213 @@
...
+using Reko.Typing;
+
+namespace Reko.Analysis
+{
+    /// <summary>
+    /// Try to rewrite indirect call statements to Applications using
+    /// user-defined data (e.g. global variables, parameters of procedures).
+    /// </summary>
+    class IndirectCallRewriter : ExpressionTypeAscender
+    {
```
uxmal: We want a unit test fixture for this new class. We need to validate that indirect calls where there is no type information available at all don't cause crashes.

src/Decompiler/Analysis/IndirectCallRewriter.cs
```diff
@@ -0,0 +1,213 @@
...
+                    {
+                        RewriteCall(stm, ci);
+                    }
+                    catch (Exception ex)
+                    {
+                        Debug.Print(
+                            "Exception {0}" + Environment.NewLine +
+                            "when rewriting indirect call statement: {1}",
+                            ex, stm);
+                    }
```
uxmal: This should be a logged error - we exceptions to be reported by users if they encounter them. Use the pattern of passing in an `IServiceProvider` interface to the constructor.

src/Decompiler/Analysis/IndirectCallRewriter.cs
```diff
@@ -0,0 +1,213 @@
...
+                Where(usedId => usedId.Storage.Equals(storage)).
+                FirstOrDefault();
+        }
+
+        private Identifier FindUsedId(CallInstruction call, Storage storage)
+        {
+            var locStorage = storage as StackLocalStorage;
+            // $HACK add return address size to stack offset to correct
+            // parameters binding. Is it correct?
+            storage =
```
uxmal: Alas, it wont be for STDAPI virtual calls. It's probably a good idea to issue a warning here, if there is no source of metadata about the indirect call. I'm slowly trying to unify the `FunctionType` and `ProcedureSignature` classes, and information about who cleans up the stack -- and how much there is on the stack -- will be available there.

src/Decompiler/Typing/ExpressionTypeAscender.cs
```diff
@@ -147,8 +147,11 @@ public DataType VisitBinaryExpression(BinaryExpression binExp)
             }
             else
                 throw NYI(binExp);
-            binExp.TypeVariable.DataType = dt;
-            binExp.TypeVariable.OriginalDataType = dt;
+            if (binExp.TypeVariable != null)
+            {
```
uxmal: I'm a concerned about these null checks. Setting the `TypeVariable` property is done as part of the "core" type analysis which happens much later in the current process, but here obviously we want to do it almost immediately after SSA transformation. 

More and more, it seems to me that `Expression` should lose the `TypeVariable` property and move it into the `TypeStore`. Then these statements could be a simple:
`typestore.SetDataTypes(binExp, dt, dt);` and you wouldn't need the the null tests. 
It's tough to reason about this when I'm on vacation though -- I will have to think about this some more on my way back to Stockholm.

src/Decompiler/Typing/ExpressionTypeAscender.cs
```diff
@@ -274,8 +278,10 @@ public DataType VisitFieldAccess(FieldAccess acc)
             throw new NotImplementedException();
         }
 
-        public DataType VisitIdentifier(Identifier id)
+        public virtual DataType VisitIdentifier(Identifier id)
         {
```
uxmal: Was there any particular reason why this became virtual?

uxmal:
This is some nice work Pavel. I'm unfortunately away from my computer again and won't be able to look at this closer until next week. As far as the `analysis-development` branch goes, your code should plug in the same way.

src/Decompiler/Typing/ExpressionTypeAscender.cs
ptomin: I have overridden it in `IndirectCallRewriter` class.

src/Decompiler/Analysis/IndirectCallRewriter.cs
ptomin: Done in a4e0da6

src/Decompiler/Analysis/IndirectCallRewriter.cs
ptomin: Done in 52344b0

src/Decompiler/Analysis/IndirectCallRewriter.cs
ptomin: STDAPI? What do you mean? The passing of arguments in `stdcall` the same as in `cdecl` (the only difference is clearing stack)

src/Decompiler/Typing/ExpressionTypeAscender.cs
ptomin: I have idea to add `storeTypes` parameter to ExpressionTypeAscender constructor. If it is 'false' then data types should not be stored in `TypeVariable`. It will be better than `null` checks

src/Decompiler/Analysis/IndirectCallRewriter.cs
ptomin: I think my solution is correct but it's "hack" so it should be replaced with clear solution.
ApplicationBuilder returns stack arguments shifted by return address size (it is always 4 on win32-x86). This code returns stack arguments positions to correct values. My attempt to fix ApplicationBuilder caused unit test failures

src/Decompiler/Typing/ExpressionTypeAscender.cs
uxmal: I'd make `ExpressionTypeAscender` abstract, move it to `Reko.Core` and then have two subclasses, one for `TypeCollector` and one for `DataFlowAnalysis`

src/Decompiler/Typing/ExpressionTypeAscender.cs
ptomin: ... but I could try to avoid making it virtual

src/Decompiler/Typing/ExpressionTypeAscender.cs
ptomin: > move it to Reko.Core and then have two subclasses, one for TypeCollector and one for DataFlowAnalysis

What do you think about changes done in 6e464ec?

src/Decompiler/Typing/ExpressionTypeAscender.cs
ptomin: New implementation does not require making `VisitIdentifier` virtual. See 93dd31e

ptomin:
Summary
1. Unit tests (see https://github.com/uxmal/reko/pull/280#discussion_r72609284). Done in 52344b0
2. Error log (see https://github.com/uxmal/reko/pull/280#discussion_r72609411). Done in a4e0da6
3. Stack parameters binding (see https://github.com/uxmal/reko/pull/280#discussion_r72609754). I think it's another issue.
4. `ExpressionTypeAscender` (see https://github.com/uxmal/reko/pull/280#discussion_r72610374). Done in 2e73c87
5. Avoid making VisitIdentifier virtual (see https://github.com/uxmal/reko/pull/280#discussion_r72610577). Done in 93dd31e

Something else, John? Or maybe it's time to merge it

uxmal:
Regarding 3: please add to the //$TODO comment a reference to the `analysis-development` branch. Once that branch is complete this current PR will make dealing with MIPS ELF binaries a lot nicer.

ptomin:
What do you mean? Replace
```
// $HACK add return address size to stack offset to correct
// parameters binding. Is it correct?
```
with
```
// $HACK add return address size to stack offset to correct
// parameters binding. Is it correct?
// Once analysis-development branch it will will make dealing with MIPS ELF binaries
// a lot nicer.
```

uxmal:
Yes. You shouldn't have to duplicate the code that Reko already has for generation of function application s. Revise the $HACK comment,  commit it, and I'll pull the PR.

ptomin:
Done in b0d6008

uxmal:
@ptomin: I've merged the indirect call rewriter code into the `analysis-development` branch.

ptomin:
Good!

ptomin:
The next step is solving problem mentioned in `$HACK` comment
@uxmal What stack offsets should be returned by `ProcedureSerializer` for `void fn(int a, int b)` signature on `win32` platform and `x86-protected` architecture? I think it should be 4 for `a` and 8 for `b` and 0 for return address. But actually it's 0 for `a` and 4 for `b

uxmal:
It seems the `ProcedureSignature` you are creating in `IndirectCallRewriter.RewriteCall` has a `ReturnAddressOnStack` property of 0, while protected x86 code typically has 4 bytes of return address on the stack. Knowing the effect a procedure has on the stack is quite a low level aspect, but it seems that `FunctionType` needs this information. More and more, the need to unify `ProcedureSignature` and `FunctionType` is becoming pressing.

Given this info, you could hack around the problem. I do believe the unification of the two classes is the better move. I will create an experimental `unify-procsig-functype` and see if the unification can be done "cheaply".

ptomin:
John, what abot my question?
> What stack offsets should be returned by ProcedureSerializer for void fn(int a, int b) signature on win32 platform and x86-protected architecture?

The problem is in stack offsets, not in `ReturnAddressOnStack` property. See code of ApplicationBuilder.VisitStackArgumentStorage:
> return frame.EnsureStackVariable(
                    stack.StackOffset - (site.StackDepthOnEntry + sigCallee.ReturnAddressOnStack),
                    stack.DataType);
But return address is already included to `site.StackDepthOnEntry`. So the correct version of it should be:
> return frame.EnsureStackVariable(
                    stack.StackOffset - site.StackDepthOnEntry,
                    stack.DataType);

uxmal:
It seems you're right. What happens when you make the change above and try running it on all the unit tests / regression suite?

ptomin:
All unit and regression tests run successfully

uxmal:
Make a PR and submit that change. I'm surprised this hasn't been noticed sooner. I will continue with the unification work ; I'm almost done.

ptomin:
See #285

ptomin:
I have created issue #286 to track it.

PR #284, opened 2016-07-31
Fix SlibReadRealModeIntServices unit test

ptomin:


xor2003:
Why you need such test? I might add tomorrow more Int. The test will be broken again. Maybe ">" (at least
" is better?

uxmal:
I agree the test is inappropriate now that the realmodeintservices.xml file is being worked on. I will revise it tomorrow when I have the chance to look at exactly what is being tested.

Issue #286, opened 2016-08-06
Problems with offsets of stack variables


ptomin:
There are problems with offsets of stack variables.
`ApplicationBuilder` returns stack arguments shifted by return address size. To work around it I have add hack to IndirectCallRewriter (see b0d6008). I think the reason is incorrect stack offsets returned by `ProcedureSerializer`.
What stack offsets should be returned by `ProcedureSerializer` for `void fn(int a, int b)` signature on `win32` platform and `x86-protected` architecture? I think it should be 4 for `a` and 8 for `b` and 0 for return address. But actually it's 0 for `a` and 4 for `b`
So `ApplicationBuilder`, `ProcedureSerializer`, `ArgumentDeserializer` and others should be modified so that they work correctly with stack ofsets

ptomin:
John, please review unit test in bdc2361. Is it correct? I think, yes. And it failed.

ptomin:
I have created branch https://github.com/uxmal/reko/compare/master...ptomin:stack-offsets
- All unit tests run successfully now
- Output for snowman-74 was changed. I think it's right changes.
- It's failed on pySample
- Outputs for other regression tests  are the same

ptomin:
Now `stack-offsets` branch is merged with `master`. So it could be closed

PR #287, opened 2016-08-06
Create unit tests to verify problems with stack offsets

ptomin:
I have created unit test for `ArgumentDeserializer` and modify one more test to verify one of the problem mentioned in #286. I have marked them as `FailedTest` to avoid failure of Travis CI

### DIFF
--- a/src/UnitTests/Core/Serialization/ArgumentDeserializerTests.cs
+++ b/src/UnitTests/Core/Serialization/ArgumentDeserializerTests.cs
@@ -19,6 +19,7 @@
 #endregion
 
 using NUnit.Framework;
+using Rhino.Mocks;
 using Reko.Arch.X86;
 using Reko.Core;
 using Reko.Core.Expressions;
@@ -39,19 +40,22 @@ namespace Reko.UnitTests.Core.Serialization
         private ProcedureSerializer sigser;
         private ArgumentDeserializer argser;
         private MsdosPlatform platform;
+        private MockRepository mr;
 
         [SetUp]
         public void Setup()
         {
+            mr = new MockRepository();
             var sc = new ServiceContainer();
             sc.AddService<IFileSystemService>(new FileSystemServiceImpl());
             arch = new X86ArchitectureReal();
             platform = new MsdosPlatform(sc, arch);
-            sigser = new X86ProcedureSerializer(
+            sigser = mr.Stub<ProcedureSerializer>(
                 arch,
                 new TypeLibraryDeserializer(platform, true, new TypeLibrary()),
                 "stdapi");
-            argser = new ArgumentDeserializer(sigser, arch, arch.CreateFrame(), 4);
+            argser = new ArgumentDeserializer(sigser, arch, arch.CreateFrame(), 13);
+            mr.ReplayAll();
         }
 
 
@@ -95,6 +99,7 @@ namespace Reko.UnitTests.Core.Serialization
         }
 
         [Test]
+        [Category(Categories.FailedTests)]
         public void ArgSer_DeserializeStackVariable()
         {
             var arg = new Argument_v1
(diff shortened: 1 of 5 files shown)


ptomin:
John, also please review my fix for `ApplicationBuilder` test (see 40edda1)

ptomin:
... and my fix for X86ProcedureSerializerTests (see fa4e0a2)

uxmal:
Pavel, I'm holding off on this PR until I've written up a document on how stacks work in Reko. There is a lot of "evolutionary crap" left from previous versions of the decompiler that probably need to be cleared out.

ptomin:
But could you tell, is my common idea of this PR right?

src/UnitTests/Core/Serialization/ArgumentDeserializerTests.cs
```diff
@@ -103,8 +108,35 @@ public void ArgSer_DeserializeStackVariable()
                 Type = new PointerType_v1 { DataType = new PrimitiveType_v1 { ByteSize = 1, Domain = Domain.Character } }
             };
             var id = argser.Deserialize(arg);
-            Assert.AreEqual("ptrArg04", id.Name);
+            Assert.AreEqual("ptrArg0D", id.Name);
             Assert.AreEqual("(ptr char)", id.DataType.ToString());
```
uxmal: Pochemu +0x0D here? Stack aligned pointers should have offsets divisible by 4.

uxmal:
This is looking good, Pavel. However, I saw some odd offsets (+0x0D) in  ArgumentSerializerTests.cs where I would have expected +0x08. Why the change there? Typically parameters have word-aligned addresses.

src/UnitTests/Core/Serialization/ArgumentDeserializerTests.cs
```diff
@@ -39,19 +40,22 @@ public class ArgumentDeserializerTests
...
             sc.AddService<IFileSystemService>(new FileSystemServiceImpl());
             arch = new X86ArchitectureReal();
             platform = new MsdosPlatform(sc, arch);
-            sigser = new X86ProcedureSerializer(
+            sigser = mr.Stub<ProcedureSerializer>(
                 arch,
                 new TypeLibraryDeserializer(platform, true, new TypeLibrary()),
                 "stdapi");
-            argser = new ArgumentDeserializer(sigser, arch, arch.CreateFrame(), 4);
+            argser = new ArgumentDeserializer(sigser, arch, arch.CreateFrame(), 13);
```
ptomin: > Pochemu +0x0D here? Stack aligned pointers should have offsets divisible by 4.
I set return address size to 13. So first argument should be at 0x0D(=13) offset

ptomin: Why I set it to 13? I think, it would be nice for unit tests to cover such strange cases

src/UnitTests/Core/Serialization/ArgumentDeserializerTests.cs
ptomin: See https://github.com/uxmal/reko/pull/287/files#r73803425

src/UnitTests/Core/Serialization/ArgumentDeserializerTests.cs
uxmal: OK, just wanted to know the justification. Since it is a bit odd, could you add a comment why the magic 13?

ptomin: ArgumentDeserializer does not know that stack aligned pointers should have offsets divisible by 4. It should just allocate variables on stack starting from return address size. Is not it?

ptomin: > could you add a comment why the magic 13?

There is no special reason to set it 13. I just set random value. It could be also 15 or 666

ptomin: > could you add a comment why the magic 13

Oh, you mean I should add comment to code and commit it. Is not it?

uxmal: Yes, since it such an unusual value. Just add a comment like
// It's possible that old, hand-written assembler passes arguments on unaligned offsets 
so that it's clear that the 0xD is not a mistake.

ptomin: Done in 23b2772

uxmal:
These changes all look good. Anything else before I pull?

ptomin:
Nothing else (except fix for it in https://github.com/uxmal/reko/compare/master...ptomin:stack-offsets branch). Now it's time to merge it

uxmal:
OK, I'll hold off until you've merged it, run all unit/regression tests.

ptomin:
I have open #289. So it could be closed now

PR #292, opened 2016-08-09
Add possibility to run regressionTests.py on linux

ptomin:
Add "mono" to command start so that it will be started on linux


uxmal:
Thanks for this.  It opens up the possibility of adding the regression tests to the Travis CI  script.

PR #293, opened 2016-08-10
Regression tests: replace absolute paths with relative

ptomin:
There are several tests with absolute paths. So they don't work on my machine.
Replace absolute paths with relative to make regression tests work on
any machine


ptomin:
Now I can get the same output on my machine as your output (except regression.log)

Issue #228, opened 2016-05-07
Data flow analysis binds user-defined procedure arguments incorrectly


ptomin:
I have created sample to reproduce it (9e113d0)
it outputs `test1(ebp, dwArg00, argc, argv)` but it should be something like ﻿`test1(argv[0], argc, "test123", 1.0)`
If CallRewriter is modified so that it creates ApplicationBuilder with ensureVariables flag set to false then it works fine. But many unit test are failed in this case.

ptomin:
I think, the reason is that the CallSite.StackDepthOnEntry is not set before running CalRewriter. So there are 2 possible fixes:
1. Calculate CallSite.StackDepthOnEntry before running CalRewriter
2. Replace calls to applications for user defined procedures at the scanning phase in the same way as external procedures

John, what do you think about it?

uxmal:
If  the `BlockWorkItem` knows the signature of a user procedure when it encounters a `RtlCall`, it should be generated as a `Application` immediately rather than a `CallStatement`. I'm surprised this isn't happening today. Similarly, `Scanner.EnsureProcedure` doesn't seem to take into account `User.Procedures` when creating the procedures.

uxmal:
Looks like `RegisterLiveness` is prepared for the changes above. I wonder why I never modified `BlockWorkItem` to deal with this. 

Pavel, is this something you feel comfortable working with or would you like me to jump in and do it? I strongly recommend having unit tests written first before attacking it.

uxmal:
Never mind! I already started on this. Keep an eye on this space....

uxmal:
Pavel: I've created a new branch `issue-228` to track this, along with a solution. Let me know if this works correctly for you.

uxmal:
By the way: is there a PR for 9e113d0 so I can incorporate it into the project?

ptomin:
> By the way: is there a PR for 9e113d0 so I can incorporate it into the project?

I have just created #233

uxmal:
Branch `issue-228` fails on the VCExeSample. 
```
l00401000: error: An error occurred during data flow analysis.
    An error occurred while processing the statement test1(dwLoc14_28, dwLoc10_23, dwLoc0C_20, dwLoc08_18).
       at Reko.Analysis.ValuePropagator.Transform(Statement stm) in D:\dev\uxmal\reko\master\src\Decompiler\Analysis\ValuePropagator.cs:line 81
[...]

PR #297, opened 2016-08-18
Create varargs format scanner

ptomin:
Create scanner for varargs format (new `VarargsFormatScanner` class). It scans and parses format and then builds application instruction with new signature. It uses children of new `IVarargsFormatParser` inerface (currently only `MsPrintfFormatParser`) to parse format
Also I have modified `MsPrintfFormatParser` so that it parses "f" as double rather than float.

### DIFF
--- a/src/Core/Types/FunctionType.cs
+++ b/src/Core/Types/FunctionType.cs
@@ -104,7 +104,7 @@ namespace Reko.Core.Types
             var sig = (FunctionType)Clone();
             sig.Parameters = Parameters.
                 Where(a => a.Name != "...").
-                Union(parameters).
+                Concat(parameters).
                 ToArray();
             return sig;
         }
--- a/src/Core/Types/FunctionType.cs
+++ b/src/Core/Types/FunctionType.cs
@@ -90,6 +90,25 @@ namespace Reko.Core.Types
             return ft;
 		}
 
+        public bool IsVarargs()
+        {
+            var last = Parameters != null ? Parameters.LastOrDefault() : null;
+            return last != null && last.Name == "...";
+        }
+
+        public FunctionType ReplaceVarargs(params Identifier[] parameters)
+        {
+            if (!IsVarargs())
+                throw new NotSupportedException(
+                    "Signature should contain varargs");
+            var sig = (FunctionType)Clone();
+            sig.Parameters = Parameters.
+                Where(a => a.Name != "...").
+                Union(parameters).
+                ToArray();
+            return sig;
+        }
+
         /// <summary>
         /// The size of the return address if pushed on stack.
         /// </summary>
--- a/src/Decompiler/Scanning/VarargsFormatScanner.cs
+++ b/src/Decompiler/Scanning/VarargsFormatScanner.cs
@@ -157,6 +157,7 @@ namespace Reko.Scanning
                 false,
                 platform.Architecture.WordWidth.Size,
                 platform.GetByteSizeFromCBasicType(CBasicType.Long),
+                platform.GetByteSizeFromCBasicType(CBasicType.Double),
                 platform.PointerType.Size);
             varargsParser.Parse();
             return varargsParser.ArgumentTypes;
--- a/src/Decompiler/Scanning/VarargsFormatScanner.cs
+++ b/src/Decompiler/Scanning/VarargsFormatScanner.cs
@@ -1,0 +1,188 @@
+﻿#region License
+/* 
+ * Copyright (C) 1999-2016 Pavel Tomin.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ */
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Reko.Core;
+using Reko.Core.CLanguage;
+using Reko.Core.Code;
+using Reko.Core.Expressions;
+using Reko.Core.Serialization;
+using Reko.Core.Types;
+using Reko.Evaluation;
+
+namespace Reko.Scanning
+{
+    /// <summary>
+    /// Try to read varargs format, then to parse it, get arguments types and
+    /// build application instruction.
+    /// </summary>
+    class VarargsFormatScanner
+    {
+        private Program program;
+        private IProcessorArchitecture arch;
+        private Frame frame;
+        private ExpressionSimplifier eval;
+        private FunctionType sig;
+
+        public VarargsFormatScanner(
+            Program program,
+            Frame frame,
+            EvaluationContext ctx)
+        {
+            this.program = program;
+            this.arch = program.Architecture;
+            this.frame = frame;
+            this.eval = new ExpressionSimplifier(ctx);
+        }
+
+        public Instruction BuildInstruction(Expression callee, CallSite site)
+        {
+            var pc = callee as ProcedureConstant;
+            if (pc != null)
+                pc.Procedure.Signature = this.sig;
+            var ab = CreateApplicationBuilder(callee, this.sig, site);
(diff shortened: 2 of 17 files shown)


src/Core/Types/FunctionType.cs
```diff
@@ -90,6 +90,25 @@ public override DataType Clone()
...
+        public FunctionType ReplaceVarargs(params Identifier[] parameters)
+        {
+            if (!IsVarargs())
+                throw new NotSupportedException(
+                    "Signature should contain varargs");
+            var sig = (FunctionType)Clone();
+            sig.Parameters = Parameters.
+                Where(a => a.Name != "...").
+                Union(parameters).
+                ToArray();
```
uxmal: Since parameter lists are ordered collections, you're better off using the `.Concat()` extension method rather than `.Union()`. The latter is fully justified in converting `(a, a).Union(a, a)` into `(a)` which isn't correct.

src/Decompiler/Scanning/VarargsFormatScanner.cs
```diff
@@ -0,0 +1,189 @@
...
+using Reko.Evaluation;
+
+namespace Reko.Scanning
+{
+    /// <summary>
+    /// Try to read varargs format, then to parse it, get arguments types and
+    /// build application instruction.
+    /// </summary>
+    class VarargsFormatScanner
+    {
```
uxmal: Classes should be `public` unless there is a strongly compelling reason for them not to be.

src/Decompiler/Scanning/VarargsFormatScanner.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        private string ReadVarargsFormat(FunctionType sig)
+        {
+            var formatIndex = sig.Parameters.Length - 2;
+            if (formatIndex < 0)
+                throw new ApplicationException(
+                    string.Format("Varargs: should be at least 2 parameters"));
+            var formatParam = sig.Parameters[formatIndex];
+            var stackStorage = formatParam.Storage as StackStorage;
```
uxmal: On non-x86 architectures, like Sparc or PowerPC, you can have varargs functions where the first `N` parameters are passed in registers and the remaining are passed on the stack. For now, leave this in, but you'll want to handle `printf` on RISC machines too.

subjects/PE-x86/VCExeSample/VCExeSample.h
```diff
@@ -373,44 +395,49 @@ typedef struct Globals {
...
+	Eq_59 gbl_c;	// 403018
 } Eq_1;
 
-typedef void (Eq_5)(char *, int32, char *, real32);
+typedef void (Eq_5)(char *, int32, char *, Eq_10);
+
+typedef union Eq_10 {
+	real32 u0;
+	real64 u1;
+} Eq_10;
```
uxmal: Yeah, this union is caused by the printf format string interpreter coercing the real32 into a real64.

subjects/PE-x86/VCExeSample/VCExeSample.c
```diff
@@ -10,9 +10,9 @@ int32 main(int32 argc, char * * argv)
 	return 0x00;
 }
 
-void test1(char * arg1, int32 arg2, char * arg3, real32 arg4)
+void test1(char * arg1, int32 arg2, char * arg3, Eq_10 arg4)
 {
-	printf("%s %d %s %f");
+	printf("%s %d %s %f", arg1, arg2, arg3, arg4);
 	return;
```
uxmal: We need to take into account that when calling a varargs function, real32's are converted to real64's. That would avoid creating the union type `Eq_10` above.

For now, we live with it, but I'd probably prefer seeing:
`printf("%s %d %s %f", arg1, arg2, arg3, (double) arg4);`

src/Decompiler/Scanning/VarargsFormatScanner.cs
```diff
@@ -0,0 +1,189 @@
...
+            EvaluationContext ctx)
+        {
+            this.program = program;
+            this.arch = program.Architecture;
+            this.frame = frame;
+            this.eval = new ExpressionSimplifier(ctx);
+        }
+
+        public Instruction BuildInstruction(Expression callee, CallSite site)
+        {
```
uxmal: I'd love to see some unit tests for `BuildInstruction` that exercise it thoroughly.

uxmal:
@ptomin: I thank you for this very significant contribution. In fact, I'm considering making a new 0.6.1.0 release with this new vargargs feature. Once you've addressed the code review comments I made, we should work on determine what remains to be done before making a release.

src/Decompiler/Scanning/VarargsFormatScanner.cs
uxmal: What this means is a `//$TODO` and a (disabled) unit test that simulates varargs on a RISC machine, where the first, say, 3 parameters are passed in registers.

src/Core/Types/FunctionType.cs
ptomin: Done in 5a95513

src/Decompiler/Scanning/VarargsFormatScanner.cs
ptomin: Done in 4cf2308

src/Decompiler/Scanning/VarargsFormatScanner.cs
ptomin: I'll try to add them at the end of week.

subjects/PE-x86/VCExeSample/VCExeSample.c
ptomin: Few months ago you have fixed x86 rewriter so that it returns cast for `fstp` if operands sizes are different (see #228). I think the same fix should be applied to `fld` and `Eq_10` will go away

ptomin: The varargs is not specific case for it. The same problem could be reproduced on any procedures with double formal arguments and float actuals

ptomin:
John, I have rebased this branch to avoid merge conflict

src/Decompiler/Scanning/VarargsFormatScanner.cs
ptomin: I have created unit tests in 818bfc8. Is it enough?

src/Decompiler/Scanning/VarargsFormatScanner.cs
ptomin: I have added `$TODO` comment in 95908f5.
Also I have created disabled `Vafs_PpcPrintf` unit test for `PowerPC` in 6e6436e. I'm not good at `PowerPC` architecture so please check if this test is correct.

subjects/PE-x86/VCExeSample/VCExeSample.c
```diff
@@ -12,7 +12,7 @@ int32 main(int32 argc, char * * argv)
 
 void test1(char * arg1, int32 arg2, char * arg3, real32 arg4)
 {
-	printf("%s %d %s %f");
+	printf("%s %d %s %f", arg1, arg2, arg3, (real64) arg4);
```
ptomin: @uxmal Looks better, is not it?

uxmal: very nice!

src/UnitTests/Scanning/VarargsFormatScannerTests.cs
```diff
@@ -0,0 +1,213 @@
...
+using Reko.Core.Expressions;
+using Reko.Core.Serialization;
+using Reko.Core.Types;
+using Reko.Scanning;
+using Reko.UnitTests.Mocks;
+
+namespace Reko.UnitTests.Scanning
+{
+    [TestFixture]
+    class VarargsFormatScannerTests
```
uxmal: `public` class here

src/UnitTests/Scanning/VarargsFormatScannerTests.cs
```diff
@@ -0,0 +1,213 @@
...
+            var appl = (Application)((SideEffect)instr).Expression;
+            var sig = ((ProcedureConstant)appl.Procedure).Procedure.Signature;
+            Assert.AreEqual(
+                "(fn void ((ptr char), (ptr char), char))",
+                sig.ToString());
+        }
+
+        [Test]
+        [Ignore("Varargs scanning has not implemented on PowerPc")]
+        public void Vafs_PpcPrintf()
```
uxmal: I can take a look at this later. The calling convention typically is dictated primarily by the `Platform`, since in theory (and certainly in practice) an assembler language programmer can pass arguments any way she wants.

src/UnitTests/Scanning/VarargsFormatScannerTests.cs
uxmal: Never mind, I'll add this myself so you don't have to.

PR #301, opened 2016-08-22
Move varargs parsers to separate namespace

ptomin:
- Create separate namespace `Reko.Core.Varargs`
- Move `printf` parser to `Reko.Core.Varargs`. It's cross-platform so it would be better than `Reko.Environments.Windows` 
- Move tests for `printf` parser to `UnitTests` assembly.

### DIFF
--- a/src/Core/Analysis/IVarargsFormatParser.cs
+++ b/src/Core/Analysis/IVarargsFormatParser.cs
@@ -21,7 +21,7 @@
 using System.Collections.Generic;
 using Reko.Core.Types;
 
-namespace Reko.Core
+namespace Reko.Core.Analysis
 {
     public interface IVarargsFormatParser
     {
--- a/src/Core/Analysis/PrintfFormatParser.cs
+++ b/src/Core/Analysis/PrintfFormatParser.cs
@@ -24,6 +24,9 @@ using System.Collections.Generic;
 
 namespace Reko.Core.Analysis
 {
+    /// <summary>
+    /// Parses standard C printf format
+    /// </summary>
     public class PrintfFormatParser : IVarargsFormatParser
     {
         private int i;
--- a/src/Core/Analysis/PrintfFormatParser.cs
+++ b/src/Core/Analysis/PrintfFormatParser.cs
@@ -18,16 +18,13 @@
  */
 #endregion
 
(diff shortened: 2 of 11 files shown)


src/Core/Varargs/IVarargsFormatParser.cs
```diff
@@ -21,7 +21,7 @@
 using System.Collections.Generic;
 using Reko.Core.Types;
 
-namespace Reko.Core
+namespace Reko.Core.Varargs
 {
```
uxmal: I suggest calling this namespace `Reko.Core.Analysis`, so that other shared analysis classes can go here.

src/Core/Varargs/PrintfFormatParser.cs
```diff
@@ -18,16 +18,13 @@
  */
 #endregion
```
uxmal: The Microsoft `wsprintf` function obeys a slightly different syntax from standard C `printf`. I don't have the exact difference at hand, but it makes sense for `Windows.MsPrintfFormatParser` to inherit from `Core.Analysis.PrintfFormatParser`.

src/Core/Varargs/IVarargsFormatParser.cs
ptomin: New 7dec332 commit overwrites this branch with change namespace to `Reko.Core.Analysis`

src/Core/Varargs/PrintfFormatParser.cs
ptomin: Do you mean `MsPrintfFormatParser` inherited from ` PrintfFormatParser` with empty impementation part should be created?

uxmal: Yes, make `MsPrintfFormatParser` inherit from `PrintfFormatParser`, with no members. I will refactor the  base class so that "extenders" of the printf format strict can plug in. For instance, the C standard says nothing about how `printf("%S", unicode_string)` should be handled (http://www.cplusplus.com/reference/cstdio/printf/), but when using MSVC, the

ptomin: Done in 6b39e06. Is it time to merge?

PR #303, opened 2016-08-24
Refactoring of ExpressionTypeAscender

ptomin:
Do changes noticed in https://github.com/uxmal/reko/commit/7b899cfc6cbbb4ccc814aee339df607571fba3fa#commitcomment-18738177
Many regression tests output were changed. I'm not sure if it's correct so please review it


src/tests/Typing/ExdUnion.exp
```diff
@@ -26,9 +26,9 @@ T_6: (in 0x0000000C : word32)
...
-  DataType: ptr32
-  OrigDataType: ptr32
+  DataType: (ptr int32)
+  OrigDataType: (ptr int32)
 T_8: (in Mem0[p + 0x0000000C:word32] : word32)
   Class: Eq_5
-  DataType: real32
-  OrigDataType: real32
+  DataType: (union (int32 u0) (real32 u1))
+  OrigDataType: (union (int32 u0) (real32 u1))
```
ptomin: Is it correct? The test is
```
Test(m.LoadDw(m.IAdd(p, 12)), PrimitiveType.Int32),
Test(m.LoadDw(m.IAdd(p, 12)), PrimitiveType.Real32)
```
So I think `union` is more correct than `real32`

ptomin:
I have fixed broken tests. John, what do you think about this changes?

src/tests/Typing/ExdUnion.exp
uxmal: Yeah, `p + 0x0C` is a 32-bit field that should be treated as a union.

PR #306, opened 2016-09-06
Create parser for PyArg_ParseTuple

ptomin:


uxmal:
Very nice! The `fp - CONST` expressions will be fixed up by the `SsaTransformer` in the `analysis-development` branch.

PR #308, opened 2016-09-07
Implement CloneExpression for PhiFunction

ptomin:
Implement `CloneExpression` for `PhiFunction`

### DIFF
--- a/src/Core/Expressions/PhiFunction.cs
+++ b/src/Core/Expressions/PhiFunction.cs
@@ -20,6 +20,7 @@
 
 using Reko.Core.Types;
 using System;
+using System.Linq;
 
 namespace Reko.Core.Expressions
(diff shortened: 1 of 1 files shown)


uxmal:
Have you run the regression tests as well? Were there any differences?

ptomin:
No differences, off course. `Travis` should discover any difference in regression tests output

ptomin:
The only difference is that one `PySample` error has gone away (but output files are the same)

uxmal:
I'm just surprised that there weren't more changes, considering that the exception this PR fixes was occurring a lot in the regression tests. Never mind then; pulling....

ptomin:
It should affect `regression.log` but I could not push my version of this file because I  have different locale

PR #309, opened 2016-09-07
Modify TypeVariable unification

ptomin:
Modify `TypeVariable` unification: `TypeVariable` should be already unified with this `DataType` by
`ExpressionTypeAscender` so just return `DataType`


uxmal:
Sorry for the delay @ptomin , "real life" has swamped me this past week.

PR #314, opened 2016-09-19
moar libc func signatures

Lukas-Dresel:
only changes for now, don't know if any changes to regressions are necessary yet


Lukas-Dresel:
Regression tests and unit tests seem unaffected

uxmal:
Thanks for this! Moar APIs, Moar processor architectures, moar moar moar!

PR #319, opened 2016-09-25
VCExeSample: create one more function

ptomin:
Output contains references to `esp` register


ptomin:
`MergeDataFlow` method of `TrashedRegisterFinder` sets value of `esp` to `Invalid`. `TrashedRegisterFinder` runs before `ssa` so it does not know about indirect call signature.  I do not know what to do with it. Maybe new version in `analysis-development` branch could work with it.

uxmal:
Understood; you may want to investigate the recent addition of `IPlatform.CreateTrashedRegisters()` would allow the `TrashedRegisterFinder` to filter out `esp` and friends from "hell nodes" (i.e. calls to non-constant destinations).

uxmal:
I.e. in `TrashedRegisterFinder.VisitCallInstruction`, at the place where it reads `//$TODO: get trash information from signature?`, make a call `program.Platform.CreateTrashedRegisters()` and then assume those are the only registers that are affected. As usual, if the call has other live-out registers, users should be able to specify this themselves.

ptomin:
I do not undestand how modification of `TrashedRegisterFinder.VisitCallInstruction` could fix it.
The problem is different stack position at different `if-then-else` branches.
e.g there are two branches first one does not contain any class and its stack position at the end is `fp - 04`. Second one contains `stdcall` but reko does not know that it's `stdcall` at this stage and so its stack position is `fp - 08`. And different stack positions caused 'Invalid' stack posistion

ptomin:
`stdcall` functions adjust stack but stack delta could not be defined without signature

uxmal:
Ah, if it's `stdcall`, all bets are off. This is a situation where the user has to provide the actual signature, or Reko needs to discover the signature before the indirect call.

ptomin:
dbefda8 fixes `VCExeSample` output but affects other regression tests

PR #322, opened 2016-09-27
Fix stack delta deserialization on X86

ptomin:
- Fix calculation of stack delta for `stdcall` and `__thiscall` calling
conventions
- Create unit tests to verify it.
- Ran regression tests
- Remove `msvcrt2.xml` entry from `reko.config`

### DIFF
--- a/src/UnitTests/Arch/Intel/X86ProcedureSerializerTests.cs
+++ b/src/UnitTests/Arch/Intel/X86ProcedureSerializerTests.cs
@@ -227,6 +227,7 @@ namespace Reko.UnitTests.Arch.Intel
             Assert.AreEqual(2, sig.Parameters.Length);
             Assert.AreEqual("this", sig.Parameters[0].ToString());
             Assert.AreEqual("ecx", sig.Parameters[0].Storage.ToString());
+            Assert.AreEqual(8, sig.StackDelta);
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


src/UnitTests/Arch/Intel/X86ProcedureSerializerTests.cs
```diff
@@ -271,6 +272,53 @@ public void ProcSer_Load_stdcall()
...
+                    }
+                }
+            };
+            Given_ProcedureSerializer(ssig.Convention);
+            mr.ReplayAll();
+
+            var sig = ser.Deserialize(ssig, arch.CreateFrame());
+            Assert.AreEqual(8, sig.StackDelta);
+            Assert.AreEqual(4, ((StackArgumentStorage)sig.Parameters[0].Storage).StackOffset);
+            Assert.AreEqual(8, sig.StackDelta);
```
uxmal: line 298 duplicates 296 and 297 has no side effects.

ptomin: Fixed in 70bffd5

ptomin: > 297 has no side effects

What do you mean?

ptomin: Ah, I see. You mean that line 298 could be removed which was done in 70bffd5

PR #333, opened 2016-10-14
Fix DfaReg00316 unit test

ptomin:
Output for `DfaReg00316` was changed. `ecx_11`, `esp_9` and `SZO_12` were disappeared from indirect call defined ids list. Is it correct, John?


uxmal:
Well, ecx is being trashed by the procedure, so we still need it in the trashed list. I have a pile of changes on my machine that may fix this, but they also add 10 to the number of failing unit tests.  It's been a busy week in the Real World, I hope to look at this on Sunday if time allows. I really want to wrap up the `analysis-development` branch so I can start leveraging the changes that were made in it.

ptomin:
John, I have removed `*.exp` file from this PR. Now it contains only the same fix for `RunTest` as in `master`. Could it be merged now?

PR #338, opened 2016-10-24
ExpressionEmitter: fix ISub

ptomin:
Fix `ISub` to avoid cast exception inside `Constant.Create` if data type of left operand is not primitive.

### DIFF
--- a/src/Core/Expressions/Constant.cs
+++ b/src/Core/Expressions/Constant.cs
@@ -383,6 +383,11 @@ namespace Reko.Core.Expressions
             return new ConstantUInt64(PrimitiveType.Word64, n);
         }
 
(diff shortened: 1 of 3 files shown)


uxmal:
@ptomin did you consider to add a `public static Constant CreateWord(int byteSize)` to the `Constant` class to do the same thing?

ptomin:
Do you mean creation `public static Constant CreateWord(int byteSize, long value)` instead of the `Word` method of `ExpressionEmitter`? Will be done soon

ptomin:
Maybe better `public static Constant Word(int byteSize, long value)` (the same style as `Word32`, `Word16`, `Byte` and so on)

PR #343, opened 2016-10-26
TrashedRegisterFinder: reset trashed registers after hell node

ptomin:
- Reset trashed registers after hell node in the same way as in
`ExpressionPropagator`
- Add stack pointer to trashed registers on Win32, Win16 and MSDos. This is the same change as in analysis branch


ptomin:
This is the fix for #341

PR #345, opened 2016-10-30
Simplify phi loops

ptomin:
This is the fix for #344.
Statements like `x=phi(y, y, x, x)` could be simplified to `x = y`


src/UnitTests/Analysis/ValuePropagationTests.cs
```diff
@@ -471,7 +472,43 @@ public void VpPhiWithConstants()
             Assert.AreEqual("r3 = 0x4711", instr.ToString());
         }
 
-		private class DpbMock : ProcedureBuilder
+        [Test(Description = "(x = phi(<exp>, <exp>, x, x)) ==> (x = <exp>)")]
+        public void VpPhiLoops()
```
uxmal: Isn't the description a little misleading? What you're doing here is making sure that if `x = phi(a_1, a_2, .. a_n)` and all the `a_n`, after value propagation, are equal, then replace the original expression with `x = a_n`.

src/Decompiler/Evaluation/ExpressionSimplifier.cs
```diff
@@ -553,6 +559,25 @@ public virtual Expression VisitPhiFunction(PhiFunction pc)
             }
         }
 
+        private Expression SimplifyPhiArg(Expression arg)
```
uxmal: Isn't this is a specialized simplification, where affine expressions of the type `a <binop> b` are being replaced in the phi arguments? Could you not use the `VisitBinaryExpression` method to perform the same simplification? If there is a specific reason why you cannot use `VisitBinaryExpression` you should document it with a comment here.

src/UnitTests/Analysis/ValuePropagationTests.cs
ptomin: `all the a_n, after value propagation, are equal`
Not all, just part of them. And the other identifiers could be simplified to `x`

src/Decompiler/Evaluation/ExpressionSimplifier.cs
ptomin: >Could you not use the VisitBinaryExpression method to perform the same simplification?

No, it could not simplify following statements:
```
y = x - 4
a = y + 4
x = phi(a, b)
```
to
```
y = x - 4
a = x
x = phi(a, b)
```
`IdBinIdc` class processes `y` as 'used in phi' and prevents propagation

Issue #353, opened 2016-11-13
Reko does not respect the little-endian byte order flag in the ELF header for MIPS


Lukas-Dresel:
When trying to load a reversing binary from tuctf ([binary](https://github.com/team-bitskrieg/CTF-writeups/tree/master/tuctf/RE%20for%2050%20plz), [writeup](https://github.com/team-bitskrieg/CTF-writeups/tree/master/tuctf/RE%20for%2050%20plz)) in reko i get many errors about not being able to rewrite instruction 'illegal'. 

This happens for examplebecause reko attempts to decode the instruction at VA 0x004013a8.
Read in little endian the opcode stored in the file there is 0x3c1c004b which is `lui gp, 0x4b`
However, reko reads 0x4b001c3c which is invalid.

It turns out that the MipsDisassembler is instantiated [here](https://github.com/uxmal/reko/blob/master/src/Core/Program.cs#L287) from the architecture the loader of the binary set. 

However, the ELF loader in always sets the architecture to big-endian [here](https://github.com/uxmal/reko/blob/master/src/ImageLoaders/Elf/ElfLoader.cs#L148)

This clashes with the instruction opcodes that are now being read out in big-endian when the file format indicates that the opcodes are instead encoded in little endian.

Lukas-Dresel:
I would attempt a fix, but I am not sure if the solution of setting the Architecture to MipsLe32Architecture is a valid solution because I'm not sure how other things related to endian-ness are then handled.

uxmal:
It's father's day here so my family won't let me at the keyboard today. However if you're willing to try the following:
1. In the class `ElfLoader` add a `endianness` parameter in the constructor. In the derived classes `ElfLoader32` and `ElfLoader64` add the `endianness` too.
2. In the class `ElfImageLoader` when creating the above derived classes, pass the endianness  which was read from the ELF header.
3. In the class `ElfLoader32` when creating the processor architecture,  use the endianness to choose between `mips-le-32` and `mips-be-32`

Reko should be able to deal with endianness after that.

Lukas-Dresel:
It currently has no support for a ProcedureSerializer for mips-le-32, should the one for big endian work?

uxmal:
The procedure serializer used by mips-be-32 should be reusable by mips-le-32.

Lukas-Dresel:
Apart from the fact that some instructions are unimplemented, e.g. rdhwr it looks good so far now.

uxmal:
[...]

PR #354, opened 2016-11-13
Fix for issue #353

Lukas-Dresel:
Correctly get MIPS architecture from machine value AND endianness (Issue #353)


uxmal:
The code changes look good, but it looks as if you haven't changed the corresponding tests in the `UnitTests` project. Try building that project and make sure all unit tests pass.

uxmal:
...and we have a green bar! Thanks for this contribution!

PR #360, opened 2016-11-14
implemented store conditional instructions, concretized unit tests

Lukas-Dresel:
fixed MipsDisassembler bug

Take a look at the unit tests, these currently fail, however that is because we do not have a Mips64Architecture and have to split the failing instructions off from the MIPS32 instructions.

Similarly take a look at the last couple unit tests, can we figure out a way to make these special registers (cc0, FCSR, ...) also accessible for these kinds of checks?

Whether you wanna merge this or not despite the failing test cases is up to you, we can even merge everything but them and then figure out how to solve this after.

### DIFF
--- a/src/Arch/Mips/MipsDisassembler.cs
+++ b/src/Arch/Mips/MipsDisassembler.cs
@@ -327,13 +327,15 @@ namespace Reko.Arch.Mips
             new AOpRec(Opcode.ld, "R2,El"),
 
             new Version6OpRec(
-                new AOpRec(Opcode.sc, "R2,El"),
+                new AOpRec(Opcode.sc, "R2,Ew"),
                 new AOpRec(Opcode.illegal, "")),
             new AOpRec(Opcode.swc1, "F2,Ew"),
             null,
             null,
 
-            new AOpRec(Opcode.scd, "R2,El"),
+            new Version6OpRec(
+                new AOpRec(Opcode.scd, "R2,El"),
+                new AOpRec(Opcode.illegal, "")),
             null, 
             null, 
             new AOpRec(Opcode.sd, "R2,El"),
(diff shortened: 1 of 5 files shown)


uxmal:
> Take a look at the unit tests, these currently fail, however that is because we do not have a Mips64Architecture and have to split the failing instructions off from the MIPS32 instructions.

I'm working on an Mips64Architecture, monitor this project...

> Similarly take a look at the last couple unit tests, can we figure out a way to make these special registers (cc0, FCSR, ...) also accessible for these kinds of checks?

I'm not sure if I'm understanding you here. You can compare against `Registers.FCSR` and `cc0`..`cc7` easily enough. Could you perhaps clarify?

Lukas-Dresel:
Well the way I'm doing it in VerifyRegisterOperand is by arch.GetRegister for access via the name. When browsing through IntelliSense I could not figure out a good way to access these registers, I must have overlooked that, I checked in the arch.

uxmal:
OK, you can continue to use `arch.GetRegister` by modifying this line in the `Registers` class:
```C#
mpNameToReg = generalRegs.Concat(fpuRegs).ToDictionary(r => r.Name);
```
changes to 
```C#
mpNameToReg = generalRegs
        .Concat(fpuRegs)
        .Concat(fpuCtrlRegs)
        .Concat(ccRegs)
        .ToDictionary(r => r.Name);

Lukas-Dresel:
I have reworked the test to use the registers directly instead, that should be better than using the names. I only used the names because I for some reason missed the Registers.* version.

Lukas-Dresel:
I'm closing this and reopening 2 other pull requests, one for everything that currently builds to get that merged in as quickly as possible and I'll have the currently failing test-cases in a seperate PR.

PR #362, opened 2016-11-15
implemented ldc1 and muld instructions

Lukas-Dresel:


uxmal:
I've pulled this into the `mips64` branch for now. Consider making PR's against `mips64` until it gets merged into `master`.

PR #364, opened 2016-11-15
added sc and scd disassembler and rewriter + rewriter unit tests

Lukas-Dresel:
As said, this is only the sc and scd changes + rewriter unit tests.


uxmal:
This was also pulled into the `mips64` branch.

PR #366, opened 2016-11-15
Added handling for repnz ret similar to repz ret

Lukas-Dresel:


uxmal:
The build is failing because of changes in a couple of x86 binaries in the regression tests. Please run the `subjects/regressionTests.py` on your machine, check the differences, and if they seem reasonable, commit + push them out.

uxmal:
@Lukas-Dresel : was this closed by accident?

PR #370, opened 2016-11-20
fixes for running tests under mono/monodevelop

a2intl:
* add capstone linux shared libs .so's (built for x86_64), and copies to test directory in UnitTest.csproj
* fixes a case-incorrect directory (MsDos -> Msdos)
* fixes an xml-encoded carriage-return to be (correctly) platform-dependent
* fixes a unit test where mono serializes xml namespace attributes in a different order


src/UnitTests/Gui/Windows/Forms/MainFormInteractorTests.cs
```diff
@@ -246,7 +247,10 @@ public void Mfi_Save()
     </user>
   </input>
 </project>";
-            Assert.AreEqual(s, interactor.Test_SavedProjectXml);
+            // xml namespaces appear in different order on different platforms, so remove
```
uxmal: The way we've solved this in the past is to use the `XmlnsHidingWriter` which strips the xmlns attributes in a derived class of `XmlWriter`. However, the `MainFormInteractor` doesn't expose a way to override the creation of an `XmlWriter` -- yet. I will accept this for now, and then rewrite `MainFormInteractor` to start using `XmlTextWriter` explicitly.

uxmal:
Thanks for this very valuable contribution! Can you confirm that all unit tests labelled `[Category(Categories.Capstone)]` are completing successfully?

a2intl:
Yes, all enabled unit tests, including all those with `[Category(Categories.Capstone)]` are passing. I'm still working on getting the `subjects/regressionTests.py` working under mono. 

For future reference, I built the libGee.External.Capstone.Proxy.so based on the issue https://github.com/9ee1/Capstone.NET/issues/1 and gist https://gist.github.com/mistressofjellyfish/5bc715c292a75e5eae3d . It looks like they still need to get that in the Capstone.NET codebase and release.

uxmal:
You way wish to give @9ee1 a poke to roll it into his project. As for me, I'm just happy someone found the time to complete this work.

Once you've gotten all the regression tests running, you can remove `capstone` from line 10 of `$(reko)/.travis.yml` and see what happens. I'm not sure if the VM the Travis-CI tests run in is an x86 or a x86-64. If the former, it might be necessary to provide an x86 version of libGee.External.Capstone.Proxy.so.

a2intl:
travis seems to be running on x86_64, all unit tests pass (now not excluding Capstone category tests): https://travis-ci.org/a2intl/reko#L3823

uxmal:
Again, thanks for this valuable contribution. It's great to have all the Capstone tests running in Travis-CI again, and have Reko ARM support on Mono. Much appreciated!

PR #373, opened 2016-11-27
Add definitions to msvcrt.inc

ptomin:
Add `floor`, `scanf`, `fopen`, `fprintf`, `fscanf`, `fread`, `fclose`


src/UnitTests/Core/Serialization/SerializedLibraryTests.cs
```diff
@@ -73,7 +73,7 @@ public void SlibReadMsvcrtXml()
 			{
 				lib = (SerializedLibrary) ser.Deserialize(stm);
 			}
-			Assert.AreEqual(26, lib.Procedures.Count);
+			Assert.AreEqual(33, lib.Procedures.Count);
 		}
```
uxmal: we really need to get rid of this test -- it's intent is not to count the number of procedure signatures in the file. I will take care of it once I get access to my development machine.

PR #378, opened 2016-12-12
WineSpec: Added Support for fixed-size integer types

MeFisto94:
Hey Guys,
Some Wine Specs (e.g. the ntoskrnl.exe.spec) contain fixed-size types such as `int64` instead of `long`, `qword` etc.

This Pull-Request adds support for those types to the WineSpecLoader

### DIFF
--- a/src/Environments/Windows/WineSpecFileLoader.cs
+++ b/src/Environments/Windows/WineSpecFileLoader.cs
@@ -195,6 +195,24 @@ namespace Reko.Environments.Windows
             case "wstr":
                 type = PointerType_v1.Create(PrimitiveType_v1.WChar16(), 4);
                 break;
+			case "uint16":
+				type = PrimitiveType_v1.UInt16();
+				break;
+			case "uint32":
+				type = PrimitiveType_v1.UInt32();
+				break;
+			case "uint64":
+				type = PrimitiveType_v1.UInt64();
+				break;
+			case "int16":
+				type = PrimitiveType_v1.Int16();
+				break;
+			case "int32":
+				type = PrimitiveType_v1.Int32();
+				break;
+			case "int64":
+				type = PrimitiveType_v1.Int64();
+				break;
             default: throw new Exception("Unknown: " + tok.Value);
             }
             args.Add(new Argument_v1 { Type = type });


uxmal:
Thanks for the contribution! To satisfy my curiosity; would you mind telling me how you discovered Reko's `spec` file capability? It's not exactly well documented, but certainly an intentional feature. Also, how would you like to be credited in the `AUTHORS` file?

MeFisto94:
Sorry for the late reply!
I found it by googling something I can't remember (maybe `reko extract library calls`). It lead me to an issue where you linked the Wine repository along with some information about that.

Should reko support `varargs`? I added them to the calling convention files (I took the StackVariable class, I guess), but it didn't analyze those parameters properly. I assume one has to add some handling code to it? On the other hand I've read somewhere that `varargs` should already work (but not in the spec files probably).

You can mention me simply as `MeFisto94` in the `AUTHORS` file, but I don't think that you have to do it, since my contribution was small. I actually only did what the Open Source Spirit is about: Send in the changes you did anyway.

uxmal:
Reko does support `varargs`, although it is a little rough around the edges and has only been tested rigorously on x86-32. Typically, a function with `varargs` parameters also supplies some extra-linguistic data, called *characteristics* in the Reko codebase, where it specifies the class name of a C# class that implements `IVarargsFormatParser`. Such classes that know how to parse the `varargs` parameters based on the parameters passed before. For instance, the `*printf` format string specifies that identity of the varargs is parsed by the class `PrintfFormatParser`.

If the wine spec files can indicate varargs (I guess one could look at the declaration of `wsprintf` in the Wine source tree) then the Wine spec file loader  should generate a vararg "placeholder" argument, naming it `...` 

Characteristics are extra-linguistic because they can't really be expressed in a C/C++-style declaration. They are stored in a separate characteristics file. For instance, Reko comes with a `win32characteristics.xml` file which contains extra-linguistic for a few Win32 api's.

MeFisto94:
Well unfortunately there are no characteristics directly in the WineSpec but there is something like this:
https://github.com/wine-mirror/wine/blob/47cf3fe36d4f5a2f83c0d48ee763c256cd6010c5/dlls/ntoskrnl.exe/ntoskrnl.exe.spec#L108

They simply specify it as calling convention. Is Reko able to know what args are still part of the call?
If not, doesn't that lead to trouble on the stack?

uxmal:
Typically varargs functions will clean up the stack after the call -- certainly the standard cdecl convention specifies this. Reko will not be derailed by a call to such a function. However, without any extralinguistic information about how many arguments were actually passed,  Reko will generate a function call where only the 'known" arguments are present.  This means that the varargs will be left out completely in the call; which in turn means that any values passed as varargs become dead.  The dead values will be incorrectly removed by Reko since it doesn't know they are live.

Providing extralinguistic support for a varargs function is not terribly hard. Perhaps you could open a separate issue where you provide the function you need this for and we can work on a solution.

PR #379, opened 2016-12-12
WineSpecLoader: Fixes loading of Metadata for non-DLL files

MeFisto94:
Since there are some libraries which aren't dlls (ntoskrnl.exe, portcls.sys) the old method of defining the Module Name has failed on those.

Instead of blindly appending .dll we now check for the extension. If there is an extension, then it's no dll file and we should pick whatever extension was defined.


uxmal:
Thanks for this!

PR #380, opened 2016-12-12
Don't crash the Application when a Spec File is invalid.

MeFisto94:
Without this, the Application would crash when you specify an invalid Wine Spec File.
Instead it would be better to just warn the user.

Note: It might be desirable to remove the MetadataFile from the Project again to enable re-adding the file when the errors have been fixed. Since I wasn't sure on how this should be implemented, I didn't do that.

Note2: For some reason my new git client messes things up (UTF-8 and Intendation), sorry for that

### DIFF
--- a/src/Gui/Forms/MainFormInteractor.cs
+++ b/src/Gui/Forms/MainFormInteractor.cs
@@ -1,6 +1,6 @@
 #region License
 /* 
- * Copyright (C) 1999-2016 John Källén.
+ * Copyright (C) 1999-2016 John Kï¿½llï¿½n.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


uxmal:
I can clean up the UTF-8 mess; don't worry about it. Will look into how hard it would be to back out the changes `ProjectLoader` makes to `Projects` when it loads them; I agree that if a file fails to load it should not affect the state of the project.

MeFisto94:
Maybe we could change this to be a warning and not an Exception that way only faulty methods would be skipped and the parsing wouldn't stop completely.

If I got that right, Reko keeps the imports registered before the faulty line? That rules out what I had in mind (removing the whole MetadataFile upon Exception).

Maybe we could check all the MetadataFiles upon loading of a new one and discard any modules with the same name. That way you have the partial-complete imports inbetween edits to the files.

uxmal:
The parsing and the building of the metadata file are horribly intermixed right now, and should really be done as separate phases, or at least be adjusted so that instead of giving up with an exception, instead emit a diagnostic but continue on (as you suggest). I'll work out a solution and see if it works better.

BTW: I asked in another PR thread how you wished to be credited in the `AUTHORS` file; let me know so I can acknowledge your contribution.

uxmal:
The Wine spec parser now filters out bad lines, with one diagnostic per bad line. This should be good enough for now, I think.

PR #382, opened 2016-12-21
Add one more test case to VCExeSample

ptomin:
- Add one more test case to `VCExeSample`. It failed with `Resulting type is (union (word32 u1) (uipr64 u0)), which isn't supported yet.` error
- I have modified `ValuePropagator` so that it continues work after this kind of errors and allows Reko to finish decompilation.
- Fix broken tests (Memory identifiers were not visited) 

The error is still present and I'm going to fix it. But now it's not interrupt decompilation at data flow analysis stage


uxmal:
I'm going to merge this into `master`, but then I will refactor the code so that we don't throw an exception in the `IdConstantRule` -- it should just warn the user. It's not a good idea to throw exceptions if they are not, well, exceptional.

PR #385, opened 2016-12-25
Create method to check if lists of uses of identifiers are correct

ptomin:
- Create `CheckUses` method to check if lists of uses of identifiers are correct and use it in `SsaTransform` and `OutParameterTransformer` tests
- Do some fixes to keep list of uses of identifiers in correct state

### DIFF
--- a/src/Decompiler/Analysis/InstructionUseAdder.cs
+++ b/src/Decompiler/Analysis/InstructionUseAdder.cs
@@ -1,6 +1,6 @@
 #region License
 /* 
- * Copyright (C) 1999-2016 John Källén.
+ * Copyright (C) 1999-2016 John KÃ¤llÃ©n.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -64,11 +64,11 @@ namespace Reko.Analysis
         public override void VisitStore(Store store)
         {
             store.Src.Accept(this);
-            /* ecxOut should not be added to use list of statements like
-               `*ecxOut = ecx` */
+            // ecxOut should not be added to use list of statements like
+            // '*ecxOut = ecx'
             if (store.Dst is Dereference)
                 return;
-            /* do not add memory identifier to uses*/
+            // Do not add memory identifier to uses
             var access = store.Dst as MemoryAccess;
             if (access != null)
             {
--- a/src/Decompiler/Analysis/InstructionUseAdder.cs
+++ b/src/Decompiler/Analysis/InstructionUseAdder.cs
@@ -44,6 +44,13 @@ namespace Reko.Analysis
             a.Src.Accept(this);
         }
 
+        public override void VisitCallInstruction(CallInstruction ci)
+        {
+            ci.Callee.Accept(this);
+            foreach(var use in ci.Uses)
+                use.Accept(this);
+        }
+
         public override void VisitDeclaration(Declaration decl)
         {
             if (decl.Expression != null)
@@ -52,6 +59,26 @@ namespace Reko.Analysis
 
         public override void VisitDefInstruction(DefInstruction def)
         {
+        }
+
+        public override void VisitStore(Store store)
+        {
+            store.Src.Accept(this);
+            /* ecxOut should not be added to use list of statements like
+               `*ecxOut = ecx` */
+            if (store.Dst is Dereference)
+                return;
+            /* do not add memory identifier to uses*/
+            var access = store.Dst as MemoryAccess;
+            if (access != null)
+            {
+                var sa = access as SegmentedAccess;
+                if (sa != null)
+                    sa.BasePointer.Accept(this);
+                access.EffectiveAddress.Accept(this);
+                return;
+            }
+            store.Dst.Accept(this);
         }
 
         public override void VisitPhiAssignment(PhiAssignment phi)
@@ -67,6 +94,10 @@ namespace Reko.Analysis
         public override void VisitIdentifier(Identifier id)
         {
             ssaIds[id].Uses.Add(user);
+        }
+
+        public override void VisitOutArgument(OutArgument outArg)
+        {
         }
 
         #endregion
--- a/src/Decompiler/Analysis/InstructionUseVisitorBase.cs
+++ b/src/Decompiler/Analysis/InstructionUseVisitorBase.cs
@@ -1,0 +1,96 @@
+﻿#region License
+/* 
+ * Copyright (C) 1999-2016 Pavel Tomin.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ */
+#endregion
+
+using Reko.Core.Code;
+using Reko.Core.Expressions;
+
+namespace Reko.Analysis
+{
+    public abstract class InstructionUseVisitorBase : InstructionVisitorBase
+    {
+
+        #region InstructionVisitor Members
+
+        public override void VisitAssignment(Assignment a)
+        {
+            a.Src.Accept(this);
+        }
+
+        public override void VisitCallInstruction(CallInstruction ci)
+        {
+            ci.Callee.Accept(this);
+            foreach (var use in ci.Uses)
+                use.Accept(this);
+        }
+
+        public override void VisitDeclaration(Declaration decl)
+        {
+            if (decl.Expression != null)
+                decl.Expression.Accept(this);
+        }
+
+        public override void VisitDefInstruction(DefInstruction def)
+        {
+        }
+
+        public override void VisitStore(Store store)
+        {
+            store.Src.Accept(this);
+            // ecxOut should not be added to use list of statements like
+            // '*ecxOut = ecx'
+            if (store.Dst is Dereference)
+                return;
+            // Do not add memory identifier to uses
+            var access = store.Dst as MemoryAccess;
+            if (access != null)
+            {
+                var sa = access as SegmentedAccess;
+                if (sa != null)
+                    sa.BasePointer.Accept(this);
+                access.EffectiveAddress.Accept(this);
+                return;
+            }
+            store.Dst.Accept(this);
+        }
+
+        public override void VisitPhiAssignment(PhiAssignment phi)
+
+        {
+            phi.Src.Accept(this);
+        }
+
+        #endregion
+
+        #region IExpressionVisitor Members
+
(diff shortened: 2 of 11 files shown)


src/Decompiler/Analysis/InstructionUseAdder.cs
```diff
@@ -54,6 +61,26 @@ public override void VisitDefInstruction(DefInstruction def)
         {
         }
 
+        public override void VisitStore(Store store)
+        {
+            store.Src.Accept(this);
+            /* ecxOut should not be added to use list of statements like
+               `*ecxOut = ecx` */
+            if (store.Dst is Dereference)
```
uxmal: Style: consider using `//` comments over C-style `/* */` comments, to stay in line with the rest of the codebase.

src/UnitTests/Analysis/OutParameterTransformerTests.cs
```diff
@@ -69,6 +69,11 @@ private void PerformTest(FileUnitTester fut)
 
 				proc.Write(false, fut.TextWriter);
 				fut.TextWriter.WriteLine("====================");
+                //$REVIEW: OutpMutual test failed. OutParameterTransformer
+                // removes phi assignments from uses. Is it correct, John?*/
+                if (proc.Name == "fn0C00_0004")
```
uxmal: This looks like debugging code that was accidentally committed.

uxmal: The phi assignment `x_2 = φ(x_0,x_1)` should indeed be a use of both `x_0` and `x_1`.

src/Decompiler/Analysis/SsaState.cs
```diff
@@ -110,12 +110,67 @@ public void Dump(bool trace)
...
-		/// <param name="pstm"></param>
- 		public void DeleteStatement(Statement stm)
+        [Conditional("DEBUG")]
+        public void CheckUses()
+        {
+            CheckUses(s => Debug.WriteLine(s));
+        }
+
+        public void CheckUses(Action<string> error)
+        {
```
uxmal: Although I like the idea of this method in principle, I'm concerned about its implementation. Is these any way you can perform the check without actually modifying the SSA state? Ideally, checks / validations should be read-only operations, with no side effects on the object being validated.

src/UnitTests/Analysis/OutParameterTransformerTests.cs
ptomin: No, it's a hack. As mentioned `OutpMutual` test failed without it. See `TransformPhiAssignment` method of  `OutParameterTransformer`. It remove phi assignment from uses. Removing this line from code could broke something else so I just added this hack. It should be removed when `OutParameterTransformer` will be fixed, off course

src/Decompiler/Analysis/SsaState.cs
ptomin: OK, I'll create new class `InstructionUssCollector` to get actually used identifiers without modifying SSA state. And I'll use it instead of `InstructionUseAdder`/`InstructionUseRemover`

src/Decompiler/Analysis/InstructionUseAdder.cs
ptomin: Done in 3c84c7b

src/UnitTests/Analysis/OutParameterTransformerTests.cs
ptomin: I have removed this hack (549dd52) and marked `OutpMutual` as failed (857ea89)

uxmal: I just looked at the `analysis-development` branch, and it turns out the `OutParameterTransformer` class is dead. There is a  `CallRewriterTests.CrwMutual`, currently disabled. I will make sure that passes before my next commit.

src/Decompiler/Analysis/SsaState.cs
ptomin: Done in a2ce1ce

ptomin:
I fix `CheckUses` so that it does not `SsaState` modification. Is it time to merge it?

src/Decompiler/Analysis/InstructionUseVisitorBase.cs
```diff
@@ -0,0 +1,96 @@
...
+ */
+#endregion
+
+using Reko.Core.Code;
+using Reko.Core.Expressions;
+
+namespace Reko.Analysis
+{
+    public abstract class InstructionUseVisitorBase : InstructionVisitorBase
+    {
```
uxmal: What other derived classes of  `InstructionVisitorBase` are there? If there are no others, it seems they could be collapsed into a single base (following the [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) principle)

ptomin: There are more then ten classes derived from `InstructionVisitorBase`

ptomin: e. g. `CritialInstruction`, `DeadCode`

src/Decompiler/Analysis/SsaState.cs
```diff
@@ -110,12 +110,69 @@ public void Dump(bool trace)
...
+                    var id = sid.Identifier;
+                    if (use == stm)
+                        if (idMap.ContainsKey(id))
+                            idMap[id]++;
+                        else
+                            idMap[id] = 1;
+                }
+            }
+            return idMap;
+        }
```
uxmal: Could this be simplified to:
```C#
var idMap =
   (from sid in sids
    from use in sid.uses
    where use == stm 
    group new { sid.id, u } by sid.id into g
    select new { Key = g.Key, Value = g.Count() })
    .ToDictionary(de => de.Key, de => de.Value);
```
?

src/Decompiler/Analysis/SsaTransform.cs
```diff
@@ -395,14 +398,15 @@ public class VariableRenamer : InstructionTransformer
...
-			/// <param name="p">procedure to rename</param>
-			public VariableRenamer(SsaTransform ssaXform)
+            /// <summary>
+            /// Walks the dominator tree, renaming the different definitions of variables
+            /// (including phi-functions). 
+            /// </summary>
+            /// <param name="ssa">SSA identifiers</param>
+            /// <param name="p">procedure to rename</param>
+            public VariableRenamer(SsaTransform ssaXform, HashSet<Statement> newPhiStatements)
 			{
```
uxmal: Could you explain the new parameter and why it is needed?

src/UnitTests/Analysis/DataFlowAnalysisTests2.cs
```diff
@@ -229,7 +229,8 @@ void test(int32 a, int32 b)
 test_entry:
 	// succ:  l1
 l1:
-	Mem8[0x00010008:word32] = a + b
+	word32 r1_7 = a + b
+	Mem8[0x00010008:word32] = r1_7
```
uxmal: This seems like a regression, unless `r1_7` is used elsewhere?

src/Decompiler/Analysis/InstructionUseVisitorBase.cs
uxmal: I mean `InstructionUseVisitorBase` -- the one you've just introduced. It seems like introducing an abstract class with only a single derived class is a little overkill? Perhaps I've missed something?

ptomin: There two classes derived from `InstructionUseVisitorBase`: `InstructionUseAdder` and `InstructionUseCollector`

src/UnitTests/Analysis/DataFlowAnalysisTests2.cs
ptomin: Maybe it processed as output parameter in the same way as `r2_6`

ptomin: This PR contains `SsaTransform` adding use instruction changes. It marks identifiers of exit block use statements as `used`. See 531 line of `SsaTransform` (https://github.com/uxmal/reko/pull/385/commits/85de2c412b2bdc7b1fad7e06a46e6ac8c6bbcc1a#diff-c32fd9c08cbe70e692e0b2a49b8810fa)
So, now there are two uses of `r1_7` (`word32 r1_7 = a + b` and `use r1_7`) and it could not be coalesced.
I see, it does not respect user signature but it's another issue. `DataFlowAnalysis2Tests` is early version of new `ssa`. So it should be solved at `analysis-development` (if it is not already fixed)

src/Decompiler/Analysis/SsaState.cs
ptomin: Oh, I'm not good at such complicated instructions. But I'll try to do it.

ptomin: Finally, done in 3dd674c

src/Decompiler/Analysis/SsaTransform.cs
ptomin: Done in 209345d

ptomin:
I have done requested fixes. Anything else, John?

uxmal:
Thanks for your patience, Pavel. One final thing: there must have been a bug you've encountered that made you change `SsaTransform`. Was it one of the existing unit tests, or some other piece of code? If the later, could you provide a unit test that exhibits the problem? The reason why I'm asking for this is that the `SsaTransform` class has been completely rewritten in the `analysis-development` branch and I want to make sure that whatever fix you've made in `master` survives into that development branch.

ptomin:
OK. There are several fixes of `SsaTansform`. I'll switch off each fix and if there will not failed unit tests, I'll create new one.

ptomin:
First step. I have removed fix for phi statement use. All test passed. Then I have added uses check to `SsaTests`. `SsarDiamond` and `SsarDiamondFrame` become failed. So now they verify this fix.
To be continued...

ptomin:
Second step. I have removed  `fp` and `Mem<num>` identifier use removing.
There is failed tests.
See e1d95df and https://travis-ci.org/uxmal/reko/builds/187380676

ptomin:
Third step. I have remove force of using id in call instruction.
 `SsaHellNode` failed.
See e0fd544 and https://travis-ci.org/uxmal/reko/builds/187383075

ptomin:
So all `SsaTransform` fixes could be verified by unit tests now. Is it enough?

uxmal:
OK, looks good. I will merge this into `analysis-development` as soon as possible to see if anything breaks and fix it there.

ptomin:
There is also problem with `OutpMutual` test. `OutParameterTransformer` class removes `a1` and `a2` uses from `a3 = phi(a1,a2)` which is incorrect. But when I tried to remove the
 ```
ssaIds[idSrc].Uses.Remove(stmDef)
```
line from `TransformPhiAssignment` method, several unit tests failed

uxmal:
Don't worry about that. I mentioned in a separate comment thread that `OutParameterTransformer` is gone in the `analysis-development` branch, and that `CrwMutual` will catch any changes in that test.

ptomin:
OK. So what is decision  about `master`? Just make `OutpMutual` ignored or remove uses check from `OutParameterTransformerTests`?

uxmal:
Go ahead and removing the uses check for now. I want to finish the `analys-development` branch soon, and hopefully out parameters won't be an issue until then.

ptomin:
> Go ahead and removing the uses check for now

OK

> I want to finish the analys-development branch soon,

Good luck!

PR #388, opened 2016-12-30
Add uses check to CoalescerTests

ptomin:
- Add uses check to `CoalescerTests`
- Add unit tests for calls
- `Coalescer`: do not replace call uses


src/UnitTests/Analysis/CoalescerTests.cs
```diff
@@ -112,7 +112,33 @@ public void CoaSideEffectCalls()
...
+        public void CoaCallCallee()
+        {
+            var m = new ProcedureBuilder("foo");
+            var r2 = m.Register(2);
+            var r3 = m.Register(3);
+            m.Assign(r3, m.Fn(r2));
+            m.Assign(r3, m.IAdd(r3, 4));
+            m.Call(r3, 4);
+            m.Return();
+            RunFileTest(m, "Analysis/CoaCallCalee.txt");
```
uxmal: Minor nit: the file should be called `CoaCallCallee.txt`

ptomin: Fixed in 7b2339d

PR #389, opened 2017-01-04
Incorrect call conventions parsing

ptomin:
- Create unit tests to verify call conventions parsing
- `NamedDataTypeExtractor`: fix applying call convention
- Fix broken unit test

### DIFF
--- a/src/Core/CLanguage/NamedDataTypeExtractor.cs
+++ b/src/Core/CLanguage/NamedDataTypeExtractor.cs
@@ -279,7 +279,15 @@ namespace Reko.Core.CLanguage
         
         public Func<NamedDataType,NamedDataType> VisitCallConvention(CallConventionDeclarator conv)
         {
+            ApplyCallConvention(conv.Convention);
             return (nt) => conv.Declarator.Accept(this)(nt);
+        }
+
+        private void ApplyCallConvention(CTokenType convention)
+        {
+            if (callingConvention != CTokenType.None)
+                throw new FormatException(string.Format("Unexpected extra calling convention specifier '{0}'.", callingConvention));
+            callingConvention = convention;
         }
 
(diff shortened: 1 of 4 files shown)


ptomin:
Strange. `C2X_pfn_arg` did not fail on `Travis`

ptomin:
> Strange. C2X_pfn_arg did not fail on Travis

It isn't ran on `Travis`. @uxmal Shouldn't it be moved to `Reko.UnitTests.dll` assembly?

uxmal:
The reason why there are unit tests embedded in the C2Xml executable is that I was experimenting with a new way of organizing unit tests. The idea is to locate unit tests in the same assembly as the code that is being tested. This would make navigation between unit tests and the code under test easier, especially now when Reko has more than 50 projects. For instance, if we were to choose to do the same thing with `Reko.Core`, the file system would look like:
```
+ Reko
+--Core
|  +--UnitTests
:
```
However, I'm not religiously attached to this idea, and it would require a lot of effort to move all the existing unit tests that we have now. If you feel that it would be better to relocate the unit tests for C2xml into `Reko.UnitTests.dll`, I'm fine with that too.

ptomin:
> If you feel that it would be better to relocate the unit tests for C2xml into Reko.UnitTests.dll, I'm fine with that too.

I just want to put `C2Xml` under `Travis` control

uxmal:
Let me take care of that; look for a commit soon.

uxmal:
Look for _two_ commits: 6e5dd5cd3be02c84d667d1dee0b6308412af482a and 00e767242cce5456e0a87ff6fe64cf673caa4c65. Travis-CI seems happy enough.

ptomin:
I have raised #390 to test it. If it'll fail then it's OK

ptomin:
Yes, it failed. It's great. Thanks, John

Issue #387, opened 2016-12-29
6.1.0 problem with scrollbar


dhucul:
Decompile a binary
Select a long function in the project browser
Attempt to scroll down with the scrollbar
Result: freezes

ptomin:
Could you provide binary so that we could reproduce it?
> Attempt to scroll down with the scrollbar

Which scrollbar?

uxmal:
Also,since this involves the user interface, could you specify which operating environment you are running under?

dhucul:
The scrollbar for the decompiled assembly (the second windows on the right side).  Windows 10 64x anniversary edition.

ptomin:
I can not reproduce it. 
Do you catch this bug on any binary?
Could you provide binary? If licence does not allow do it, could you generate sample binary?
P.S. @dhucul Have you closed it by mistake?

Testato:
Same here, win10-64bit:

```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: Linear address 000636C7 is not in known segment.
   at Reko.Core.SegmentMap.MapLinearAddressToAddress(UInt64 linearAddress)
   at Reko.Gui.Windows.Controls.DisassemblyTextModel.SetPositionAsFraction(Int32 numerator, Int32 denominator)
   at Reko.Gui.Windows.Controls.TextView.vScroll_ValueChanged(Object sender, EventArgs e)
   at System.Windows.Forms.ScrollBar.OnValueChanged(EventArgs e)
   at System.Windows.Forms.ScrollBar.set_Value(Int32 value)
   at System.Windows.Forms.ScrollBar.DoScroll(ScrollEventType type)
   at System.Windows.Forms.ScrollBar.WmReflectScroll(Message& m)
   at System.Windows.Forms.ScrollBar.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)


************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1586.0 built by: NETFXREL2
[...]

PR #394, opened 2017-01-09
Fix problem with disassembly scrollbar

ptomin:
Create unit test to reproduce #387 
To be continued...

### DIFF
--- a/src/Gui/Windows/Controls/DisassemblyTextModel.cs
+++ b/src/Gui/Windows/Controls/DisassemblyTextModel.cs
@@ -154,10 +154,10 @@ namespace Reko.Gui.Windows.Controls
             long offset = Math.BigMul(numerator, (int)mem.Length) / denominator;
             if (offset < 0)
                 offset = 0;
-            else if (offset > mem.Bytes.Length)
-                offset = mem.Bytes.Length;
-
-            this.position = program.SegmentMap.MapLinearAddressToAddress(addrStart.ToLinear() + (uint)offset);
+            var addr = addrStart + offset;
+            if (addr >= addrEnd)
+                addr = addrEnd-1;
+            this.position = addr;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


uxmal:
I'll hold off until you've found the issue.

ptomin:
All unit tests passed. @uxmal please review and merge

src/Gui/Windows/Controls/DisassemblyTextModel.cs
```diff
@@ -154,10 +154,10 @@ public void SetPositionAsFraction(int numerator, int denominator)
...
                 offset = 0;
-            else if (offset > mem.Bytes.Length)
-                offset = mem.Bytes.Length;
-
-            this.position = program.SegmentMap.MapLinearAddressToAddress(addrStart.ToLinear() + (uint)offset);
+            var addr = addrStart + offset;
+            if (addr >= addrEnd)
+                addr = addrEnd-1;
+            this.position = addr;
         }
```
uxmal: Consider changing this to:
```
var linAddr = addrStart + offset;
if (addr >= addrEnd) {
    addr = addrEnd - 1;
// Normalize address
addr = program.SegmentMap.MapLinearAddressToAddress(addr.ToLinear());

ptomin:
> This will fail on large x86 realmode segmented binaries

What about unit test to reproduce it?

uxmal:
I'm in the middle of something else -- I guess I can accept your PR, with the warning that it won't work on msdos binaries, and then write a unit test for that once I've caught up.

ptomin:
> I'm in the middle of something else

Maybe I'll try to create such test

ptomin:
`DisassemblyTextModel` works with only one segment. So 
> treating the memory as one big array

is correct. Is not it?

uxmal:
Pavel, I'm going to pull this now and then I will work with the MS-DOS binary to see if it triggers any crashes. OK with you?

ptomin:
OK

PR #397, opened 2017-01-13
Type ascender: meet address-of data type

ptomin:
- Type ascender tests: add test for address-of expression
- Type ascender: meet address-of data type


ptomin:
@uxmal All checks have passed. Could you review it?

PR #398, opened 2017-01-14
Fix ComplexExpressionBuilder

ptomin:
- Create unit tests for `ComplexExpressionBuilder`
- Modify `ComplexExpressionBuilder`: return `<expression>` instead of `*&<expression>`

PR #399, opened 2017-01-15
ComplexExpressionBuilder: more cases to do *&id => id transform

ptomin:
Do changes requested by @uxmal (https://github.com/uxmal/reko/pull/398#issuecomment-272644503)


uxmal:
Ochen' horosho! That fix looks more general. Thanks for hunting this down.

PR #400, opened 2017-01-17
StructureMerger throws exception on structures with different sizes

ptomin:
Create unit tests for `TypeTransformer` to reproduce exception throwing in `StructureMerger` on structures with different sizes


ptomin:
I have fixed `UnionPointersStructuresMatcher`. Now it does not match structures with different sizes.

PR #401, opened 2017-01-18
SignatureGuesser throws NullReferenceException

ptomin:
Microsoft mangled name parser is really great thing. It allows to process a lot of external global variables and signatures without user help.
But `SignatureGuesser` throws exception for virtual tables.


ptomin:
I have fixed `SignatureGuesser`

PR #402, opened 2017-01-21
Self-referencing structure clone causes stack overflow

ptomin:
Create self-referencing structure clone unit test


ptomin:
I have modified data type clone. Now visited structures are stored in special dictionary

PR #403, opened 2017-01-21
Remove UnionTypeVarsReplacer

ptomin:
John, I tried to remove `UnionTypeVarsReplacer` using from `TypeStore`. All unit and regression tests passed. Is there any benefit of `UnionTypeVarsReplacer`? Or it could be removed.


uxmal:
I'm afk at the moment. There obviously was a reason for introducing it many years ago but the recent changes in the Reko type analysis may have rendered it useless.

uxmal:
I think we can go ahead with this. `TypeVariable` is used in a different way from the days of `TraitCollector`.

PR #405, opened 2017-01-21
680x0 disassembler bugfix: invalid cpBcc instructions no longer result in invalid M68kInstructions

Kalmalyzer:
Hi,

when disassembling a certain 68000 program ("p61con", The Player 6.1's module converter) I noticed that the GUI disassembler would throw an unhandled exception very quickly if I scrolled around a bit in the disassembly view window.

This change fixes a bug in one of the code paths which generate disassembled instructions. (The code that converted from disassembled instruction to string representation would trip itself up on it.) With the change in place I haven't noticed any unhandled exceptions when browsing that source code.

### DIFF
--- a/src/Arch/M68k/M68kDisassembler.cs
+++ b/src/Arch/M68k/M68kDisassembler.cs
@@ -1362,10 +1362,11 @@ namespace Reko.Arch.M68k
         {
             var new_pc = dasm.rdr.Address;
             dasm.LIMIT_CPU_TYPES(M68020_PLUS);
+            Opcode opcode = g_cpcc[dasm.instruction & 0x3f];
             return new M68kInstruction
             {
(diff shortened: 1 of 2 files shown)


uxmal:
Thank you for this contribution! Could you do me a great favor and add a comment that include the Hex bytes that the disassembler choked on? Since you're not familiar with unit tests I'll write one for you.

Kalmalyzer:
I got curious... So I have implemented a unit test for this + verified that the test fails without my fix, and works with my fix. Hope that will do?

uxmal:
Perfect! Thanks for this. How would you like to be credited in AUTHORS?

Kalmalyzer:
"Mikael Kalms" will do. Thanks :)

PR #411, opened 2017-01-29
Improve phi function simplification

ptomin:
Create one more unit test for loops simplification to reproduce #410


ptomin:
Simplification of phi functions was improved in 71c0f72

uxmal:
Great work Pavel; more expressions of the type `sp + C1 + C2` are being simplified to `fp + (C1 + C2)` now. I did see some statements of the type `ss:fp + C1 + 0x00` in pp.c that may need to be looked at, but I'm pulling this now as it improves code generation considerably.

PR #412, opened 2017-02-02
ArrayAccess and Slice matching is not implemented

ptomin:
Create unit tests for `ExpressionMatcher`


ptomin:
Fixed in a2c60e3

uxmal:
Pavel, do you want to implement the pattermatching for those two cases before I pull the request?

ptomin:
At the moment, I'm not ready to do full implementation of it

ptomin:
I think it could be merged as is

PR #413, opened 2017-02-04
IndirectCallExpander: use DefStatement rather than DefExpression to expand identifies

ptomin:
Use `DefStatement` rather than `DefExpression` to expand identifies


uxmal:
Nice improvement, Pavel. In general, I think that not depending on `SsaIdentifier.DefExpression` is a good idea. I'm considering removing it in the `analysis-branch` and replacing it with a `GetDefiningExpression()` method. The reason of course is that the `ValuePropagation` and `Coalescer` transformations will mutate assignments so that the `DefExpression` properly needs to be updated over and over again.

PR #416, opened 2017-02-05
Do not enqueue procedure from user global marked as no-decompiled

ptomin:
Create unit test for no-decompiled procedure from user global


ptomin:
Fixed in d4affa4

PR #421, opened 2017-02-12
Create unit test for `SliceGegmentedPointer` rule

ptomin:
Create unit test for `SliceGegmentedPointer` rule


uxmal:
Thanks for that. I've fixed the bug (and another one I found while fixing the first one!)

PR #422, opened 2017-02-18
Implement type inference of indirect calls

ptomin:
Implement type inference of indirect calls


uxmal:
Looks good Pavel. Could you also add a unit test to ExpressionTypeDescenderTests?

ptomin:
I have created unit test in 7892b3b

Issue #424, opened 2017-02-26
System.InvalidCastException on x86 DOS .exe


fenugrec:
Hi,
using the latest 0.6.2 msi available on sourceforge, I'm having this problem :

- load a 16-bit DOS .exe ( this one, https://filebin.net/a6f0c1gpdfxf5v5r ) ;
- click "Action -> Next phase", try to browse around
- if I select any of the automatically-named functions (say, fn0800_01B4), reko works for a few seconds and then I get this .NET exception window (reko doesn't crash however) : 

(I trimmed the detailed version information at the end)
```
************** Exception Text **************
System.InvalidCastException: Unable to cast object of type 'Reko.Gui.Windows.Controls.EmptyEditorModel' to type 'Reko.Gui.Windows.Controls.MixedCodeDataModel'.
   at Reko.Gui.Windows.CombinedCodeViewInteractor.MixedCodeDataView_TopAddressChanged()
   at Reko.Gui.Windows.CombinedCodeViewInteractor.CombinedCodeView_CurrentAddressChanged(Object sender, EventArgs e)
   at Reko.Core.EventEx.Fire(EventHandler eh, Object sender)
   at Reko.Gui.Windows.Controls.CombinedCodeView.set_CurrentAddress(Address value)
   at Reko.Gui.Windows.CombinedCodeViewInteractor.set_SelectedAddress(Address value)
   at Reko.Gui.Windows.CombinedCodeViewInteractor.DisplayProcedure(Program program, Procedure proc)
   at Reko.Gui.Windows.CodeViewerServiceImpl.DisplayProcedure(Program program, Procedure proc)
   at Reko.Gui.Design.ProcedureDesigner.DoDefaultAction()
   at Reko.Gui.ProjectBrowserService.tree_AfterSelect(Object sender, EventArgs e)
   at Reko.Core.EventEx.Fire(EventHandler eh, Object sender)
   at Reko.Gui.Windows.Controls.TreeViewWrapper.treeView_AfterSelect(Object sender, TreeViewEventArgs e)
   at System.Windows.Forms.TreeView.OnAfterSelect(TreeViewEventArgs e)
   at System.Windows.Forms.TreeView.TvnSelected(NMTREEVIEW* nmtv)
   at System.Windows.Forms.TreeView.WmNotify(Message& m)
   at System.Windows.Forms.TreeView.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
[...]

PR #425, opened 2017-02-26
Limit num procedure lines at Code View Panel of Combined Code Viewer

ptomin:
This is workaround for #424


uxmal:
I'll accept this as a workaround. Long-term we should figure out what can be done to make the CombinedCodeViewer "virtual" so that it doesn't have to render so much code .

PR #426, opened 2017-02-27
Implement ComplexExpressionBuilder::VisitVoidType

ptomin:
- Create unit test to verify `ComplexExpressionBuilder::VisitVoidType`
- `ComplexExpressionBuilder`: implement `VisitVoidType`

### DIFF
--- a/src/Decompiler/Typing/ComplexExpressionBuilder.cs
+++ b/src/Decompiler/Typing/ComplexExpressionBuilder.cs
@@ -101,6 +101,52 @@ namespace Reko.Typing
             return exp;
         }
 
+        /// <summary>
+        /// Return fallback expression. This is something like
+        ///    (char *)exp + offset + index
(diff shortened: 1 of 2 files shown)


src/Decompiler/Typing/ComplexExpressionBuilder.cs
```diff
@@ -101,6 +101,52 @@ public Expression BuildComplex(bool dereferenced)
...
+            else
+            {
+                DataType dt;
+                if (enclosingPtr != null)
+                    dt = new Pointer(PrimitiveType.Char, enclosingPtr.Size);
+                else
+                    dt = PrimitiveType.CreateWord(expComplex.DataType.Size);
+                e = new Cast(dt, expComplex);
+            }
+            if (offset != 0)
```
uxmal: This logic looks very similar to `CreateArrayIndexExpression` in this same class. Could it be reused?

PR #428, opened 2017-03-03
ComplexExpressionBuilder: use FallbackExpression() (part № 2)

ptomin:
Use FallbackExpression() when we could not build appropriate C-like expression


ptomin:
I have modified `CompexExpressionBuilder` so that it uses `FallbackExpression()`. New unit tests were created
Could you review it, John?

uxmal:
I will review on Sunday night. AFK this weekend.

PR #434, opened 2017-03-18
Create SsaProcedureBuilder and use it in ValuePropagationTests

ptomin:
Create `SsaProcedureBuilder` and use it in `ValuePropagationTests`
`SsaProcedureBuilder` supports the building of a procedure in Static Single Assignment form.
Some unit tests require procedure to be in Static Single Assignment form. This class gives possibility
to build it without `ssa` transforming

### DIFF
--- a/src/UnitTests/Analysis/ValuePropagationTests.cs
+++ b/src/UnitTests/Analysis/ValuePropagationTests.cs
@@ -41,61 +41,20 @@ namespace Reko.UnitTests.Analysis
 	[TestFixture]
 	public class ValuePropagationTests : AnalysisTestBase
 	{
-		SsaIdentifierCollection ssaIds;
         private MockRepository mr;
         private IProcessorArchitecture arch;
         private IImportResolver importResolver;
         private FakeDecompilerEventListener listener;
-        private ProcedureBuilder m;
+        private SsaProcedureBuilder m;
 
         [SetUp]
 		public void Setup()
 		{
-			ssaIds = new SsaIdentifierCollection();
             mr = new MockRepository();
             arch = mr.Stub<IProcessorArchitecture>();
             importResolver = mr.Stub<IImportResolver>();
             listener = new FakeDecompilerEventListener();
-            m = new ProcedureBuilder();
-        }
-
-        private Identifier Reg32(string name)
-        {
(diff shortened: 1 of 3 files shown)


uxmal:
@ptomin: sorry to not have responded sooner. I finally had time to look at this properly. I'm wondering if it makes sense to just rewrite the `ValuePropagatorTests` class to use `SsaTransform`. The approach of using `ProcedureBuilder` and then `SsaTransform` should be simple enough that we don't need to introduce another class. `SsaTranform` is a relatively clean transform without too many dependencies, so I think it is a more appropriate. What do you think?

ptomin:
If we will use `SsaTransform` then ValuePropagatorTests will be not **unit** tests, it will be integration tests. I think unit tests should be independent of other classes (e.g. `SsaTransform`). And `SsaTransform` changes will not cause changes to output of `ValuePropagator` unit tests.

ptomin:
And I'm going to use `SsaProcedureBuilder` for new `IndirectCallRewriter` unit test (see dd69411)

uxmal:
Two observations: `ValuePropagator` requires procedures to be in SSA form, so you have to provide procedures in SSA form, either by building the SSA form by hand as was previously done in `ValuePropagatorTests`, using this new `SsaProcedureBuilder` , or building the SSA form the way Reko does using `SsaTransform`.

I grant you that the master branch's implementation of `SsaTransform` feels "bulky", with its requirement to run `AliasAnalysis` first to detect registers that alias each other. Perhaps this is what is making you feel that it's not "unit test enough", as it has dependencies on a few other classes. I don't agree; I think it's OK to use `SsaTransform` in unit tests if `SsaTransform` doesn't depend on the class being tested. The `SsaTransform` implementation in `master` has fewer dependencies, which might make it more palatable for you to use.

The proposed `SsaProcedureBuilder` class looks very similar to how I began implementing  the new `SsaTransform` implementation in the `analysis-branch`. Were that branch ready for production, I would just go ahead and use it. Sadly, I cannot yet merge that branch into `master` because it requires some features I'm adding to `scanning-branch`.

Having said that, I acknowledge the effort you've put into this PR and that you are already using it for testing other parts of Reko. I will accept it, but don't be surprised if it later, once `analysis-branch` is merged into `master`, I will replace it with `SsaTransform`.

PR #438, opened 2017-04-23
Do not overrride user-defined stack delta

ptomin:
Improve  X86ProcedureSerializer: do not override user-defined stack delta

### DIFF
--- a/src/Arch/X86/X86ProcedureSerializer.cs
+++ b/src/Arch/X86/X86ProcedureSerializer.cs
@@ -59,6 +59,8 @@ namespace Reko.Arch.X86
                 d == "__thiscall" ||
                 d == "pascal")
                 sig.StackDelta += StackOffset;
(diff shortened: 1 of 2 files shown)


uxmal:
Nice PR, Pavel. Do you have any ideas of what the generated signature of the procedure should look like in the output c file? If the procedure `foo` is discovered violating a standard calling convention, should there be an attribute:
```
int [[reko:stack_delta(8)]] foo(int dwArg08, int dwArg0C, int dwArg10)
```
I actually have an MS-DOS executable generated by Turbo Pascal in my "zoo" that contains such procedures.

ptomin:
Yes, `[[reko:stack_delta(8)]]` is appropriate attribute

uxmal:
And if Reko can detect that the parameters comform with ```__stdcall```,  ```__cdecl``` or any other platform-defined convention, it should of course do that.

Issue #348, opened 2016-11-04
Reko works incorrectly with floating point compare statement


ptomin:
For instance, it could be reproduced on `VCExeSample`. `(d < 321.0)` statement was decompiled as `globals->r4020F0 < rArg04` (see https://github.com/uxmal/reko/commit/37ab72ddc33da38e49a179646f5944d056083b58#commitcomment-19666936 and https://github.com/uxmal/reko/commit/37ab72ddc33da38e49a179646f5944d056083b58#commitcomment-19666961) but it should be `globals->r4020F0 > rArg04`

uxmal:
The `d > 1.0` in `test7` is also inverted. Likely a bug in the code that deals with
```asm
fld1	
	fcomp double ptr [ebp+08]
	fstsw	 ax
	test ah,05
	jpe 00401189
```

uxmal:
The culprit is likely in the `MatchesFstswSequence()`  method in `https://github.com/uxmal/reko/blob/master/src/Arch/X86/X86Rewriter.Fpu.cs`. Pavel, do you want to take a stab at this?

ptomin:
Yes, I could try to fix it

ptomin:
Fixed in a20e4fd. Could be closed, now

PR #440, opened 2017-05-09
Add unit test for missed fstsw sequence

ptomin:
Following sequence was missed during working under #348
```
fstsw ax
text ah, 01
je <address>
```


uxmal:
Good catch Pavel!

PR #441, opened 2017-05-13
X86 Rewriter rewrites fdivr incorrectly

ptomin:
`fdivr arg` was rewritten to `arg = arg / st(0)`. Should be `st(0) = arg / st(0)`


uxmal:
Another bug squashed!

PR #442, opened 2017-05-14
Reko does not show warning for not found import procedure which name started with underscore

ptomin:
`Reko` does not show warning for not found import procedure which name started with underscore


uxmal:
This is a helpful addition; it may spur a user to provide his own metadata.

ptomin:
> it may spur a user to provide his own metadata.

Yes, I used it to discover new  set of undefined `CRT` procedures (see #443)

PR #444, opened 2017-05-16
xbuild is depreciated

rogersachan:
See http://www.mono-project.com/docs/about-mono/releases/5.0.0/


rogersachan:
>>>> xbuild tool is deprecated and will be removed in future updates, use msbuild instead <<<<

Right from Travis CI

PR #445, opened 2017-05-18
WIP: MSBuild support for VS for Mac/Linux (Mono)

rogersachan:
This PR introduces the use of the new MSBuild system, but cannot be merged until https://developercommunity.visualstudio.com/content/problem/54843/error-msb4019-microsoftwebapplicationtargets-was-n.html is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: csharp
 solution: src/Reko-decompiler.sln
 install:
   - nuget restore src/Reko-decompiler.sln
-  - nuget install NUnit.Runners -Version 2.6.4 -OutputDirectory testrunner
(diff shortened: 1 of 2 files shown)


uxmal:
Thanks for this.  Any ETA for VStudio for Mac?

smx-smx:
I compiled Reko pn msbuilc succesfully the other day, on a 32bit system and i was hitting the same problems with msbuild a while ago.
Maybe it's worth mentioning that the build failed on the first try, but running msbuild again (without cleaning) worked then

rogersachan:
I'm going to make a new PR for this, since the behavior for msbuild has changed since then.

PR #447, opened 2017-06-02
Do not enqueue procedure from user global marked as no-decompiled (restore fix)

ptomin:
Fix done in #416 was broken during work on scanner.


uxmal:
This must have gotten lost in the big changes that were made in the scanner. Thanks for patching this up.

PR #448, opened 2017-06-05
VCExeSample: swicth off shingle scanner

ptomin:
The reason of adding `VCExeSample` to regression tests collection was testing of decompiling of interesting functions.
Decompiling of CRT code generated by linker is not useful. So I have switched off `Shinled scanner` to decompile only part of code created by user.


uxmal:
Agreed. We've known for a while that Reko needs an option to read + create libraries of function signatures that it can use to identify statically linked runtime procedures that are not needed in the final disassembly.

PR #449, opened 2017-06-10
Add test case to VCExeSample

ptomin:
Add one more test case to `VCExeSample`


uxmal:
Nothing like a good challenge eh?  :-)

ptomin:
Is https://github.com/uxmal/reko/commit/06c5d723d70c604b2bc13807cb2354a447937a82#commitcomment-22475698 not a good challenge?

uxmal:
Да, очень трудно !   You have a knack for finding tough challenges Pavel.
Determining the presence of nested types automatically  is challenging. if
you like I can write you some thoughts I have on the matter. Right now I'm
out with the family but this evening I will be back at keyboard .

--
John Källén - +46739914646
Sent from my mobile device
Apologies for brevity and misspellings

ptomin:
> Да, очень трудно !
Нет, не очень трудно ( No, it's not very hard). One line `SignatureBuilder` change could fix the most problems. And then adding one case to `Unifier` should fix the rest problems. I have this fixes on my local machine and am going to prepare pull requests

uxmal:
Ah you will provide the subtype as user supplied data. I was thinking of the automatic detection of substructure w/o user intervention -- which is definitely трудно :-)     In any case looking forward to your PR.

ptomin:
Automatic detection of substructure without user intervention it's not difficult, it's impossible in general!
For instance there is not difference between
```
struct
{
    int a;
    int b;
    int c;
}
```
and
```
struct
{
    struct
    {
        int a;
        int b;
    } inner;
    int c;
}
```
But there are some cases which give chance to reconstruct. For example if we have
```
struct
{
    int a;
    int b;
    int c;
}
```
and
```
struct
{
    int a;
    struct
    {
        int b;
        int c;
    } inner;
}
``` 
and there is statement
```
foo(&val->inner)
```

ptomin:
> Determining the presence of nested types automatically is challenging. if you like I can write you some thoughts I have on the matter

Though automatic detection has not high priority for me, you could post your thoughts as issue, so that it will not be lost.

uxmal:
Doing impossible things is what decompilation is all about. Agreed that if the binary doesn't provide enough hints Reko will not be able to distinguish between a "flat" and a nested structure.  The address - of construct is also ambiguous :  in machine code it's hard to see if we're taking the address of val->inner and val->inner.a

uxmal:
I will write up my comments when I'm back home

ptomin:
>  it's hard to see if we're taking the address of val->inner and val->inner.a

but if there are operations with `c` field in `foo` then `Reko` could distinguish them

uxmal:
Indeed. That is the crux: if reko sees evidence that `ptr` is a pointer, and then sees this:
``` 
ptr_2 = ptr + 8
```
and later:
```
Mem42[ptr_2:word32] = 0x1234
```
It should be able to deduce `ptr_2` is pointing into the middle of whatever `ptr` is pointing to. I will write a unit test for this when I get back tonight and see if I can get it to pass.

PR #451, opened 2017-06-10
Unifier: merge fields into nested structures

ptomin:
Create unit test for nested structure unification

### DIFF
--- a/src/Core/Types/DataType.cs
+++ b/src/Core/Types/DataType.cs
@@ -89,7 +89,19 @@ namespace Reko.Core.Types
             return dt as T;
         }
 
-		protected void ThrowBadSize()
+        public T TypeReferenceAs<T>() where T : DataType
+        {
+            DataType dt = this;
+            TypeReference typeRef = dt as TypeReference;
+            while (typeRef != null)
+            {
+                dt = typeRef.Referent;
+                typeRef = dt as TypeReference;
+            }
+            return dt as T;
+        }
+
+        protected void ThrowBadSize()
 		{
 			throw new InvalidOperationException(string.Format("Can't set size of {0}.", GetType().Name));
 		}
--- a/src/Core/Types/Unifier.cs
+++ b/src/Core/Types/Unifier.cs
@@ -602,8 +602,7 @@ namespace Reko.Core.Types
                 this.unifier = unifier;
             }
 
-            public bool Match(
-                StructureField fa, StructureField fb)
+            public bool Match(StructureField fa, StructureField fb)
             {
                 this.fa = fa;
                 this.fb = fb;
@@ -613,11 +612,14 @@ namespace Reko.Core.Types
                 this.NextFieldB = null;
                 var strFa = fa.DataType.TypeReferenceAs<StructureType>();
                 var strFb = fb.DataType.TypeReferenceAs<StructureType>();
+                // only one field should be nested structure
                 if (
                     (strFa == null && strFb == null) ||
                     (strFa != null && strFb != null))
                     return false;
 
+                // check which of two fields is nested structure and store it
+                // and other field in corresponding variables.
                 int strSize;
                 if (strFa != null)
                 {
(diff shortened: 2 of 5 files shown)


ptomin:
The fix is coming soon

ptomin:
Fixed in cbb64eb

subjects/PE-x86/VCExeSample/VCExeSample.c
```diff
@@ -140,7 +140,7 @@ void nested_structs_test12(nested_structs_type * dwArg04)
 {
 	dwArg04->a = 0x01;
 	dwArg04->str->b = 0x02;
-	dwArg04->dw0008 = 0x03;
+	dwArg04->str->c = 0x03;
```
ptomin: Looks better, is not it, John?

src/Core/Types/DataType.cs
```diff
@@ -89,7 +89,19 @@ public DataType Clone()
...
+            DataType dt = this;
+            TypeReference typeRef = dt as TypeReference;
+            while (typeRef != null)
+            {
+                dt = typeRef.Referent;
+                typeRef = dt as TypeReference;
+            }
+            return dt as T;
+        }
+
```
uxmal: This looks very similar to `DataType.ResolveAs`. Was there a particular reason why you couldn't use it? It seems to me that `DataType.ResolveAs` could be improved to handle this case as well.

src/Core/Types/Unifier.cs
```diff
@@ -581,7 +588,44 @@ private bool TryMakeFieldName(StructureField fa, StructureField fb, out string n
             return true;
         }
 
-		public DataType UnifyPointer(Pointer ptrA, DataType b)
+        private bool IsNestedStructure(
+            StructureField fa, StructureField fb,
+            out StructureField fNestedStruct, out StructureField fOther)
+        {
+            var strFa = fa.DataType.TypeReferenceAs<StructureType>();
```
uxmal: I'm not too keen on multiple `out` parameters. It seems like a better name for this would be `FindNestedStructure` with either a `Tuple<StructureField,StructureField>` or a new class `NestedStructureFields` as a return type.

I'm also having a little trouble following the logic in the method. Perhaps a few judicious comments would clarify?

src/Core/Types/Unifier.cs
```diff
@@ -581,7 +588,44 @@ private bool TryMakeFieldName(StructureField fa, StructureField fb, out string n
...
+            fNestedStruct = null;
+            fOther = null;
+            if (
+                (strFa == null && strFb == null) ||
+                (strFa != null && strFb != null))
+                return false;
+
+            var strSize = ((strFa != null) ? strFa : strFb).GetInferredSize();
+            fNestedStruct = (strFa != null) ? fa : fb;
+            fOther = (strFb != null) ? fa : fb;
```
uxmal: `strFb != null` implies `strFa == null` (see line 599). Why not change this to 
`fOther = (strFa != null) ? fb : fa;` or even better:
```
if (strFa != null) {
    fNestedStruct = fa;
    fOther = fb;
} else {
    fNestedStruct = fb;
   fOther = fa;
}
var strSize = fNestedStruct.GetInferredSize();
```
Although a little longer, the logic is clearer and you avoid the repetition of the comparison three times.

src/Core/Types/Unifier.cs
ptomin: What about moving `IsNestedStructure` and `MergeFieldIntoNestedStructure` to new `NestedStructureUnifier` class? And so `fNestedStruct`, `fOther` become internal fields and shouldn't be returned.

uxmal: Yeah, I'd prefer that.

ptomin: New `NestedStructureUnifier` class was created in c4277c4

src/Core/Types/Unifier.cs
ptomin: Done in 9ceffe4

src/Core/Types/Unifier.cs
ptomin: > I'm also having a little trouble following the logic in the method. Perhaps a few judicious comments would clarify?

Done in 262f602. Is it clear now?

uxmal:
Let's take it the way it is. I'm going to add another unit test to see what happens if a conflicting field is present.

PR #452, opened 2017-06-12
ComplexExpressionBuilder builds pointer to nested struct field incorrectly

ptomin:


ptomin:
Fixed in 7385e6a

src/Decompiler/Typing/ComplexExpressionBuilder.cs
```diff
@@ -425,7 +425,7 @@ private Expression CreateUnreferenced(DataType dt, Expression e)
 
         private Expression CreateFieldAccess(DataType dtStructure, DataType dtField, Expression exp, Field field)
         {
-            if (enclosingPtr != null)
+            if (enclosingPtr != null && !dereferenceGenerated)
             {
```
uxmal: Nice!

PR #458, opened 2017-06-17
Move data type deserialisation from SignaturGuesser to ImportResolver

ptomin:
Only platform-defined types could be resolved by `SignatureGuesser` now.
Move data type deserialisation from `SignatureGuesser` to `ImportResolver` so that user-defined types could be resolved.


uxmal:
This makes more sense -- thanks for cleaning this up Pavel.

PR #459, opened 2017-06-18
SignatureGuesser: set stack delta for "cdecl" signature with invalid parameters

ptomin:
Stack delta is set to 0 by default. It's incorrect. Should be the same as
return address on stack for `cdecl` signatures.


uxmal:
Thanks for finding and fixing this Pavel. I'm wondering what we can do make these `stackDelta` errors harder to make.

PR #460, opened 2017-06-24
Move signature deserialisation from SignatureGuesser to ImportResolver

ptomin:
Data type deserialisation was moved from `SignatureGuesser` to `ImportResolver` in 2aedc38 (see #458).
The same thing should be done for signature deserialisation


uxmal:
Thanks for this, Pavel.

PR #472, opened 2017-08-12
Fixed crash when loading a PE image containing one or more null section names

smx-smx:


uxmal:
Ignore the appveyor error - I will fix it tomorrow

PR #475, opened 2017-08-26
CB80 output [debug]

ptomin:


ptomin:
Very strange. `Travis` run on old version shows differences of `CB80` files. But there were not such differences on the SAME version before. Maybe `Travis` build machine settings were changed

PR #478, opened 2017-09-09
Fix Travis build

ptomin:


ptomin:
Finally `Travis` build passed. Could be merged now

smx-smx:
Good, that also fixed the Appveyor build, which now fails to run the tests. It tries to use .NET 3.5 but it should be using the same version that reko is built with, i.e .NET 4.5

uxmal:
@ptomin :  ты герой!  I will review the PR in a few hours when I'm back at the keyboard

uxmal:
I will work on getting the appveyor unit tests working. Again, thanks all for your help in getting the tests back on the rails again.

PR #479, opened 2017-09-11
Fix structure analysis to make it work on unit test with break inside of nested ifs

ptomin:
Create structure analysis unit test with break inside of nested ifs. It failed


ptomin:

uxmal:
Nice work, Pavel.

PR #481, opened 2017-09-17
Make structure analysis work if there are coincident switch targets

ptomin:
Create structure analysis unit test with coincident switch targets. It failed


ptomin:
I have fixed structure analysis so that it works if there are coincident switch targets. Could be merged now

uxmal:
Thanks for this. I especially like how the PP.EXE sample output is getting cleaned up with your latest PR:s.

PR #483, opened 2017-09-23
Do not lose label during edge virtualization 

ptomin:
Do not lose label during edge virtualization if destination region has not statements


uxmal:
@ptomin: at first glance (I haven't had the time to look closely at the PR yet) it seems as if we're getting a lot of "dead" labels introduced here; it looks like a regression. Is this PR part of a series of improvements? It would be useful if you included a unit test that exhibits the issue you are trying to address.

ptomin:
> it seems as if we're getting a lot of "dead" labels introduced here
Which of introduced labels are dead? I can not find any

uxmal:
OK, I think I understand the change now. The labels were previously missing. Thanks for the PR!

ptomin:
I have added unit test in 41b7797

uxmal:
Thanks for the unit test, Pavel!

PR #486, opened 2017-10-13
Initial M68k PE loader

smx-smx:


uxmal:
looks good, don't worry about the tabs.

PR #487, opened 2017-10-14
Schwartz algorithm refines switches with irregular case exits incorrectly

ptomin:
I have created structure analysis unit test with irregular case exits. `Reko` structure analysis failed on it.
It emits
```
    switch (r1)
    {
    case 0x00:
        if (!done())
            goto l1;
        break;
    case 0x01:
case_1:
        r1 = 0x01;
        break;
    }
    return r1;
```
The correct answer is
```
    switch (r1)
    {
    case 0x00:
        if (!done())
        {
            r1 = 0x02;
            goto case_1;
        }
        break;
    case 0x01:
case_1:
        r1 = 0x01;
        break;
    }
    return r1;
```

Schwartz algorithm has not solution for this case
I'm going to implement definition of case body using the same method as for lexical nodes of loops

### DIFF
--- a/src/Decompiler/Structure/StructureAnalysis.cs
+++ b/src/Decompiler/Structure/StructureAnalysis.cs
@@ -610,7 +610,7 @@ all other cases, together they constitute a Switch[].
             Region follow = null;
             foreach (var s in regionGraph.Successors(n))
             {
-                if (regionGraph.Predecessors(s).Count != 1)
+                if (regionGraph.Predecessors(s).Where(p => (p != n)).Any())
                     return null;
                 var ss = SingleSuccessor(s);
                 if (s.Type != RegionType.Tail)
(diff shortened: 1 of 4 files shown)


ptomin:
I have improved switch refinement
- Virtualizing of all outgoing edges described by Schwartz
causes incorrect refinement of switch with irregular case
exits. Use loop lexical nodes definition method to find
case body. Then virtualize any edge leaving the case body
that does not go to the successor
- Do switch refinement after the whole graph has been
traversed
- And some other changes

Please review it, John

uxmal:
@ptomin : just chiming in to let you know I'm reviewing this PR. There were a few changes in the regression suite (most notable pp.exe) and I'm looking at them to see if they're innocuous.

uxmal:
Sorry for the delay. Thanks for this contribution!

PR #509, opened 2017-12-09
Fix structure analysis to make it work on unit test with switch inside of do-while loop

ptomin:
Create structure analysis unit test with switch inside of do-while loop. It failed


ptomin:

uxmal:
Nice work, Pavel. I especially like how the regression tests were improved by this commit.

PR #511, opened 2017-12-10
Fix structure analysis to make it work on unit test with switch with irregular entries and tail cases

ptomin:
Create structure analysis unit test with switch with irregular entries and tail cases. It failed


ptomin:

PR #513, opened 2017-12-17
Fix structure analysis to make it work on unit test with single case switch

ptomin:
Create structure analysis unit test with single case switch. It failed


ptomin:

Issue #514, opened 2017-12-17
Reko.Core.AddressCorrelatedException : Unexpected 65 fstsw mask for mov opcode .


smx-smx:
Happening in the following code snippet

```
004B684D DF E0 fstsw ax
004B684F F6 C4 41 test ah,41
004B6852 8B 44 24 04 mov eax,[esp+04]
004B6856 75 04 jnz 004B685C
```

Reko assumes that the instruction coming right after `test` is a jump of some sort, but it's not the case here because we have a `mov` in between

PR #515, opened 2017-12-17
Added unit test for issue #514

smx-smx:


uxmal:
Let's see if we can fix this.

PR #520, opened 2017-12-23
Improve last resort refinement of switches

ptomin:
- Create structure analysis unit test to reproduce error in the last resort refinement of switch
- Improve last resort refinement of switches. Schwartz algorithm suggests to first remove edges whose source does not dominate its target, nor whose target dominates its source.
Current implementation in reko removes edge from first node in the list.
It causes unnecessary gotos.


ptomin:
I have merged it. Any objections, John?

uxmal:
Looks good to me, Pavel. I like the improvements in the while-nested-in-switches in the subjects.

PR #523, opened 2017-12-27
PeImageLoader: ignore exports table if missing

smx-smx:


src/ImageLoaders/MzExe/PeImageLoader.cs
```diff
@@ -520,7 +521,8 @@ public override RelocationResults Relocate(Program program, Address addrLoad)
             ImageSymbols[entrySym.Address] = entrySym;
             var entryPoints = new List<ImageSymbol> { entrySym };
             ReadExceptionRecords(addrLoad, rvaExceptionTable, sizeExceptionTable, ImageSymbols);
-            AddExportedEntryPoints(addrLoad, SegmentMap, entryPoints);
+			if(rvaExportTable != 0)
```
uxmal: something;s wrong with the formatting here? Spaces after "if", and prefer putting single-line "then" or "else"  statements in {brackets} unless they are in { `goto`, `break`, `continue`, `return` }

PR #525, opened 2017-12-30
Fix structure analysis to make it work on unit test with "while" loop with break at the end of loop body

ptomin:
- Create structure analysis unit test with "while" loop with break at the end of loop body
- Virtualize loop break edges in different ways for "while" and "do-while"


subjects/PDP-11/lunar.c
```diff
@@ -174,23 +174,22 @@ void fn0444()
...
-		do
+		while (65356 - r0_15 <= 0x00)
 		{
-			if (65356 - r0_15 > 0x00)
-				goto l0464;
 			r0_15 = r0_15 + 0x0168;
-		} while (r0_15 > 0x00);
+			if (r0_15 > 0x00)
+				break;
+		}
```
ptomin: Condition should be inverted. But it's another issue

ptomin:
Merged. Are you agree, John?

uxmal:
Мне нравится.  I do want to look closer at the pdp-11 regression but I will be away from my computer until the evening of 1 Jan .

ptomin:
> I do want to look closer at the pdp-11 regression

Do you mean  https://github.com/uxmal/reko/pull/525#discussion_r159122198? I have fixed it on my local machine.

uxmal:
Yes. There are some other issues there as well that are not related to structure reconstruction.

ptomin:
I hope pdp-11 regression was fixed in 526

PR #526, opened 2017-12-30
Fix structure analysis to make it work on structure analysis unit test with do-while loop with inverted condition

ptomin:
See https://github.com/uxmal/reko/pull/525#discussion_r159122198
- Create structure analysis unit test with do-while loop with inverted condition
- Modify structure analysis to invert do-while condtion if necessary


ptomin:
Looks better, is not it, John. Merged

uxmal:
Отлично, Павел!    Flushed out quite a few bugs there

PR #527, opened 2017-12-30
Fix structure analysis to make it work on unit test with "do-while" loop with break at the start of loop body

ptomin:
- Create structure analysis unit test with "do-while" loop with break at the start of loop body
- Improve loop type determining during virtualizing of irregular exits

```
do
{
    process();
    if (cancel())
        break;
} while (next());
```

looks better than

```
while(true)
{
    process();
    if (cancel())
        break;
    if (!next())
        break;
}
```


ptomin:
@uxmal Semantic is the same. Style is better. Merged.

uxmal:
Хорошо еще раз!  These are great improvements Pavel.

PR #530, opened 2018-01-03
Appveyor: save the Reko msi as artifact

smx-smx:


uxmal:
Thanks @smx-smx, what a great contribution! This will make it easier for people who can't or won't build Reko themselves to work with the most recent code.

Issue #528, opened 2018-01-02
Sega Genesis environment doesn't work in .msi version


Kalmalyzer:
Hi,

I noticed that Reko gives the following error messages if I install the pre-built version and attempt to open a Sega Genesis dump:

```
Unable to open memory map file 'SegaGenesis.xml. Could not open file 'C:\Program Files (x86)\jklSoft\Reko\SegaGenesis.xml'.
An internal error occurred. Object reference not set to an instance of an object.
```

The first bit is because SegaGenesis.xml is not part of Product.wxs. If I add the following line to Product.wxs, close to where Trs80Rom.xml is listed, then the error message changes:

```
        <File Source="$(var.SegaGenesis.ProjectDir)SegaGenesis.xml" />
```
... with that line in place, the new installer-package version will instead say:
```
Unable to open memory map file 'SegaGenesis.xml. Access to the path 'C:\Program Files (x86)\jklSoft\Reko\SegaGenesis.xml' is denied.
An internal error occurred. Object reference not set to an instance of an object.
```

If I take the entire Reko installation and move it outside of C:\Program Files (x86)  then it will no longer complain. I don't know about these Windows things, sounds like something involving app manifests or configs or ... eh?

Breakpointing in the code yields:

```
        public static MemoryMap_v1 LoadMemoryMapFromFile(IServiceProvider svc, string mmapFileName, IPlatform platform)
        {
            var cfgSvc = svc.RequireService<IConfigurationService>();
            var fsSvc = svc.RequireService<IFileSystemService>();
            var diagSvc = svc.RequireService<IDiagnosticsService>();
            try
            {
                var filePath = cfgSvc.GetInstallationRelativePath(mmapFileName);
                var ser = SerializedLibrary.CreateSerializer(
                    typeof(MemoryMap_v1),
                    SerializedLibrary.Namespace_v4);
                using (var stm = fsSvc.CreateFileStream(filePath, FileMode.Open)) <===== this throws an exception
                {
                    var mmap = (MemoryMap_v1)ser.Deserialize(stm);
                    return mmap;
                }
            }
            catch (Exception ex)
            {
                diagSvc.Error(ex, string.Format("Unable to open memory map file '{0}.", mmapFileName));
                return null;
            }
        }
```
Exception:
```
-		ex	{"Access to the path 'C:\\Program Files (x86)\\jklSoft\\Reko\\SegaGenesis.xml' is denied."}	System.Exception {System.UnauthorizedAccessException}
[...]

PR #531, opened 2018-01-03
Sega genesis installer fixes

Kalmalyzer:
Fixes for the remaining problems in #528 -

1. The `SegaGenesis.xml` file was not included in the installation package. This has been fixed.
2. All memory map files were being opened in read/write mode. This would work when running a non-installer build, but fail with an access denied for an installer build (if the application was installed to `Program Files (x86)`). This has been fixed.

With these two changes, I can open a Sega Genesis binary with no errors.


uxmal:
Thanks for the contribution!

PR #534, opened 2018-01-03
Improve last resort refinement of loops

ptomin:
- Create structure analysis unit test to reproduce error in the last resort refinement of loop
- Improve last resort refinement of loops. Do the same changes as for switches in f143650


src/Decompiler/Structure/StructureAnalysis.cs
```diff
@@ -1141,11 +1141,8 @@ private bool RefineLoop(Region head, ISet<Region> loopNodes)
...
                     return true;
             }
-            foreach (var n in lexicalNodes)
-            {
-                if (LastResort(n))
-                    return true;
-            }
+            if (LastResort(lexicalNodes))
+                return true;
             return false;
```
uxmal: Perhaps this could be simplified to just:
```C#
return LastResort(lexicalNodes);
```

uxmal: Those big diffs are daunting; I'll trust that you've reviewed them and that they look good.

ptomin: > Perhaps this could be simplified to just:
Done. I have amended last commit. See https://github.com/uxmal/reko/pull/534/commits/fcdd58040cd64fac446a2b8fd12a5d5f3f11b5d1#diff-74a4adb62549845650e5bb45ba06ffff

uxmal:
Мне кажется все хорошо.  Thanks Pavel

src/Decompiler/Structure/StructureAnalysis.cs
ptomin: >Those big diffs are daunting; I'll trust that you've reviewed them and that they look good.

It's hard for me to review all of them by hand.

ptomin: The changes are like

```
label:
<some code>
goto label
```

->

```
while(...)
{
<some other code>
}
```

PR #536, opened 2018-01-04
CALLM is the appropriate mnemonic according to M68000PRM

Kalmalyzer:
The title says it all!


uxmal:
Do you have an idea of the details of the CALLM / RTM instructions?  The manual I have is very terse on the subject.

Kalmalyzer:
There is some written about CALLM/RTM in the MC68020UM, chapter 9.7 and 9.8. I'm not sure it's worth the trouble implementing them as it is doubtful any existing software contains it.

https://groups.google.com/forum/#!topic/comp.sys.amiga.programmer/LGh2fBzLkDc has some good discussion around it.

PR #537, opened 2018-01-04
M68K disassembler prints hexstrings like '$xxxx' and '-$xxxx'

Kalmalyzer:
The current M68K disassembly format for hex numbers is not acceptable as input to M68k assemblers. This change brings it closer to that ideal.

`FormatValue()`, `FormatSignedValue()` and `FormatUnsignedValue()` outside of the M68k implementation had to get parameterized in the process. It is a bit clunky; rather, I think it would be better if that parameterization was a per-architecture setting where M68k would have a format string like '{0}${1}' and x86 would have '{0}{1}h' etc.

Also, `FormatSignedValue()` as it was previously implemented would always add + signs. I think this is because `FormatSignedValue()` was being used to build expressions like `[eax+1234]` -- that is, it was used to format a value that was supposed to be chainable in an expression. This chainability is now - again, a bit clumsily, admitted - expressed via the `SignRequiredForSignedIntegers` parameter.

I think the code would become easier to follow if `FormatValue()` would have a `SignRequired` parameter and apply that logic to all types of inputs - signed integers, unsigned integers, and reals alike. That change would result in no changes in the disassembly outputs / regression test expected results, but it would need the setting to be explicitly spelled out in a number of places in the code for all other architectures.

What are other people's opinions on this?


Kalmalyzer:
Thanks, style changes implemented.

src/Core/Machine/MachineOperand.cs
```diff
@@ -56,26 +56,17 @@ public string ToString(MachineInstructionWriterOptions options)
         public abstract void Write(MachineInstructionWriter writer, MachineInstructionWriterOptions options);
 
 		/// <summary>
-		/// Should signed integers always be formatted with a leading sign character?
+		/// Converts a signed integer constant to string representatioon.
+		/// </summary>
+		/// <param name="signRequiredForSignedIntegers">sign Should signed integers always be formatted with a leading sign character?
```
ptomin: Should be `forceSign`

src/Core/Machine/MachineOperand.cs
```diff
@@ -87,7 +78,16 @@ public static string FormatSignedValue(Constant c, SignRequiredForSignedIntegers
...
+		/// <summary>
+		/// Converts a numeric constant to string representatioon.
+		/// </summary>
+		/// <param name="signRequiredForSignedIntegers">sign Should signed integers always be formatted with a leading sign character?
+		/// Setting this to Yes allows chaining a sequence of numbers into an expression,
+		///  like "+5+7-3+9".
+		/// Setting this to No will format numbers like "5" and "-3" which is normal for standalone numbers.</param>
+		/// <param name="integerFormatSpecifier">Format string; allows injecting platform-specific characters before/between/after sign and value
+		/// when printing an integer value as hex. {0} will be the sign character (if any) and {1} will be the absolute value.</param>
+		public static string FormatValue(Constant c, bool signRequiredForSignedIntegers = true, string integerFormatSpecifier = "{0}{1}")
```
ptomin: Should be forceSign

src/Core/Machine/MachineOperand.cs
```diff
@@ -87,7 +78,16 @@ public static string FormatSignedValue(Constant c, SignRequiredForSignedIntegers
 
         private static readonly char[] floatSpecials = new char[] { '.', 'e', 'E' };
 
-        public static string FormatValue(Constant c, SignRequiredForSignedIntegers signRequiredForSignedIntegers = SignRequiredForSignedIntegers.Yes, string integerFormatSpecifier = "{0}{1}")
+		/// <summary>
+		/// Converts a numeric constant to string representatioon.
+		/// </summary>
+		/// <param name="signRequiredForSignedIntegers">sign Should signed integers always be formatted with a leading sign character?
```
ptomin: Should be `forceSign`

src/Core/Machine/MachineOperand.cs
```diff
@@ -56,26 +56,17 @@ public string ToString(MachineInstructionWriterOptions options)
...
-		}
-
-		public static string FormatSignedValue(Constant c, SignRequiredForSignedIntegers signRequired = SignRequiredForSignedIntegers.Yes, string formatSpecifier = "{0}{1}")
+		/// Setting this to No will format numbers like "5" and "-3" which is normal for standalone numbers.</param>
+		/// <param name="integerFormatSpecifier">Format string; allows injecting platform-specific characters before/between/after sign and value
+		/// when printing an integer value as hex. {0} will be the sign character (if any) and {1} will be the absolute value.</param>
+		public static string FormatSignedValue(Constant c, bool signRequired = true, string formatSpecifier = "{0}{1}")
 		{
-			string s = (signRequired == SignRequiredForSignedIntegers.Yes ? "+" : "");
+			string s = (signRequired ? "+" : "");
```
ptomin: Should be `forceSign`

src/Core/Machine/MachineOperand.cs
ptomin: Maybe `format` is better than `integerFormatSpecifier`

src/Core/Machine/MachineOperand.cs
```diff
@@ -56,26 +56,17 @@ public string ToString(MachineInstructionWriterOptions options)
...
-			/// Positive integers will have no sign in front; negative will have -
-			/// </summary>
-			No
-		}
-
-		public static string FormatSignedValue(Constant c, SignRequiredForSignedIntegers signRequired = SignRequiredForSignedIntegers.Yes, string formatSpecifier = "{0}{1}")
+		/// Setting this to No will format numbers like "5" and "-3" which is normal for standalone numbers.</param>
+		/// <param name="integerFormatSpecifier">Format string; allows injecting platform-specific characters before/between/after sign and value
+		/// when printing an integer value as hex. {0} will be the sign character (if any) and {1} will be the absolute value.</param>
+		public static string FormatSignedValue(Constant c, bool signRequired = true, string formatSpecifier = "{0}{1}")
```
ptomin: Maybe `format` is better than `formatSpecifier`

src/Arch/M68k/M68kDisassembler.cs
```diff
@@ -35,6 +35,8 @@ namespace Reko.Arch.M68k
 
     public partial class M68kDisassembler : DisassemblerBase<M68kInstruction>
     {
+		public const string HexStringFormat = "{0}${1}";
+
```
ptomin: These are tabs! Should be indented by spaces

src/Arch/M68k/MemoryOperand.cs
```diff
@@ -157,13 +152,10 @@ public static MachineOperand PostIncrement(PrimitiveType dataWidth, AddressRegis
...
-            {
-                writer.Write('$');
-                writer.Write(Offset.IsNegative
-                    ? MachineOperand.FormatSignedValue(Offset)
-                    : MachineOperand.FormatUnsignedValue(Offset));
-            }
+			if (Offset != null)
+			{
+				writer.Write(MachineOperand.FormatValue(Offset, false, M68kDisassembler.HexStringFormat));
+			}
```
ptomin: Should be indented by spaces

src/Core/Machine/MachineOperand.cs
ptomin: ... and should be indented by spaces

ptomin:
Regression tests failed! Their changes should be committed.

Kalmalyzer:
Thanks! I have applied most of the changes that were suggested.

Exceptions/notes:

1. I have given `forceSign` and `format` longer, more complicated names when they are `FormatValue()` parameters. This is because the parameters only apply under some circumstances. I think having the same short names as for `FormatSignedValue()` is misleading. I believe the best resolution is, well, see item 3.

2. I have provided updates to the regression tests, they are part of this PR (they are in the first commit). The Travis CI tests will continue to fail until the PR is merged since the regression test compares against what's already stored in the Git repo. If you are not happy with this, then please let me know how I should organize the content of the PR instead.

3. Longer term, I think the right thing to do is to treat `FormatValue()` & co as if they format for two different purposes: either format a value for use as a standalone value, or format a value as part of an expression. The `signRequired` parameter should not have any default value; rather, we should go back and change all calls to `FormatValue()` & co to include an explicit value for the `signRequired` parameter. Also, the `signRequired` logic should apply for signed ints, unsigned ints, and reals. When done correctly, this will not lead to any regression changes. I'm happy to do this as long as you are OK with the plan.

ptomin:
> I have provided updates to the regression tests, they are part of this PR (they are in the first commit). The Travis CI tests will continue to fail until the PR is merged since the regression test compares against what's already stored in the Git repo. If you are not happy with this, then please let me know how I should organize the content of the PR instead.

You should run regression test on your machine by `regressionTests.py` script executing. And then commit the changes of test outputs.

Kalmalyzer:
> You should run regression test on your machine by `regressionTests.py` script executing. And then commit the changes of test outputs.

Right. That is what I have done. The changes to *.exp and *.asm files in https://github.com/uxmal/reko/pull/537/commits/9625072da8934429b53d83f8ba7f8a18659d0b09 are the result of running `regressionTests.py`. Is there anything else I should do?

ptomin:
Travis build discovers following changes
```
	modified:   Hunk-m68k/BENCHFN.asm
	modified:   Hunk-m68k/BENCHLNG.asm
	modified:   Hunk-m68k/BENCHMUL.asm
	modified:   Hunk-m68k/BYTEOPS.asm
	modified:   Hunk-m68k/FIBO.asm
	modified:   Hunk-m68k/MATRIXMU.asm
	modified:   Hunk-m68k/MAX.asm
	modified:   Hunk-m68k/STRLEN.asm
	modified:   Hunk-m68k/TESTLONG.asm
	modified:   PE/m68k/hello_m68k.asm
	modified:   regressions/reko-351/a.asm
	modified:   regressions/reko-367/code.asm
```

Try to run `regressionTests.py` one more time

Kalmalyzer:
My mistake: the regression test results were from an intermediate run. Sorry about that. Updated results, let's see if these pass.

ptomin:
Tests have passed. Good.
Now let's wait for @uxmal 's review

uxmal:
This look fine. I agree that the overloaded use of `FormatValue` is causing problems, and should be cleaned up in a separate commit. Thanks for this contribution, @Kalmalyzer!

PR #549, opened 2018-01-14
Make structure analysis work on unit test with return inside of if statement

ptomin:
- Create structure analysis unit test with return inside of if statement 
- Virtualize return edges before last resort refinement

### DIFF
--- a/src/Decompiler/Structure/Region.cs
+++ b/src/Decompiler/Structure/Region.cs
@@ -47,6 +47,17 @@ namespace Reko.Structure
         public List<AbsynStatement> Statements { get; set; }
         public Expression Expression { get; set; }
 
+        /// <summary>
+        /// Return true if region consists of a single AbsynReturn statement.
+        /// </summary>
+        public bool IsReturn
+        {
+            get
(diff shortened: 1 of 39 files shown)


src/Decompiler/Structure/Region.cs
```diff
@@ -47,6 +47,17 @@ public class Region
         public List<AbsynStatement> Statements { get; set; }
         public Expression Expression { get; set; }
 
+        /// <summary>
+        /// Check if region contains only return statement
+        /// </summary>
```
uxmal: Thanks. I'd also change the comment to make it extra clear "Return true if region consists of a single AbsynReturn statement."

ptomin: Done. I have amended this commit one more time.

uxmal:
Looks good now. Let's pull it.

uxmal:
Travis-CI builds are starting to exceed some 10 minute boundary. I woudn't be too concerned; the AppVeyor build is fine. I'll go ahead and pull now.

PR #551, opened 2018-01-20
LoadedPageInteractor -> ScannedPageInteractor

ptomin:
Actually `LoadedPageInteractor` interactor is executed after program scanning. So it should be scanned page interactor.


uxmal:
Fine by me.

PR #552, opened 2018-01-20
Fix statuses of buttons after "Finish decompilation" action 

ptomin:
- Use factory to create analyzed and final page interactors. It makes unit tests creation easier
- Create unit test to reproduce bug with statuses of buttons after `Finish decompilation` action
- Set correct current phase after `Finish decompilation` action. This is the fix for statuses of buttons after `Finish decompilation` action. `Next phase` and `Finish decompilation` should be disabled when `Finish decompilation` action is done. But it was enabled

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,8 +33,8 @@ after_build:
         mv $filePath $newPath;
 
 test_script:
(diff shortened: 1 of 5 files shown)


ptomin:
I just noticed that user interface tests could be successfully executed on `appveyor` machine. So there is no reason to exclude them

uxmal:
Thanks for this, Pavel. Looks good.

PR #553, opened 2018-01-20
Enable "Finish decompilation" button before scanning

ptomin:
- Create unit test for status of `Finish decompilation` button
- Enable `Finish decompilation`button at initial page after loading of binary. It was disabled so user should run `Scan Binary` and then press `Finish decompilation`. Now user could press only `Finish decompilation` and `Reko` will do `Scan`, `Analysis` and `Reconstruct data types`. I think, it's more convenient.


uxmal:
Agreed -- it is more convenient.

PR #559, opened 2018-01-27
Some speedups

VladRassokhin:
Subj, based on dotTrace profiling while Reko was analyzing DOS application
Also for some reason Rider messed up UTF-8 symbols in copyright comments, please  change before merging, I've failed :(

Also I've question why SortedLists used for storages. Seems ordering is not actually used in some places and lists could be replaced with HashSets which are quite faster. 

If you're interested may provide dotTrace snapshots for you, or at least in text representation.
PS: I'm not C# but Java developer so some ideas may be strange.

### DIFF
--- a/src/Core/ImageMap.cs
+++ b/src/Core/ImageMap.cs
@@ -1,6 +1,6 @@
 #region License
 /* 
- * Copyright (C) 1999-2018 John Källén.
+ * Copyright (C) 1999-2018 John Kï¿½llï¿½n.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -63,7 +63,7 @@ namespace Reko.Core
             {
                 // Outside of range.
                 Items.Add(itemNew.Address, itemNew);
-                MapChanged.Fire(this);
+                FireMapChanged();
                 return itemNew;
             }
             else
@@ -79,13 +79,13 @@ namespace Reko.Core
                         itemNew.Size = (uint)(item.Size - delta);
                         item.Size = (uint)delta;
                         Items.Add(itemNew.Address, itemNew);
-                        MapChanged.Fire(this);
+                        FireMapChanged();
                         return itemNew;
                     }
                     else
                     {
                         Items.Add(itemNew.Address, itemNew);
-                        MapChanged.Fire(this);
+                        FireMapChanged();
                         return itemNew;
                     }
                 }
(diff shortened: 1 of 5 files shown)


ptomin:
Regression tests failed. Following files differs
```
        modified:   Hunk-m68k/BENCHFN.h
	modified:   Hunk-m68k/BENCHLNG.h
	modified:   Hunk-m68k/BENCHMUL.h
	modified:   Hunk-m68k/BYTEOPS.h
	modified:   Hunk-m68k/FIBO.h
	modified:   Hunk-m68k/MAX.h
	modified:   Hunk-m68k/TESTLONG.h
	modified:   PE/m68k/hello_m68k.h
```
For more details see https://github.com/uxmal/reko/blob/master/CONTRIBUTING.md

uxmal:
Hello Vlad, and spasibo bolshoe for your contribution. Certainly we're interested in improving the performance of Reko; we already have a few issues related to the subject. Any concrete data (such as the dotTrace snapshots you mentioned) you can provide would be useful -- my focus on feature implementation sometimes leaves the performance work wanting.

Having said that, I'm reminded of the old adage: "You have a performance problem. You introduce a cache. You now have two problems." @ptomin points out that there were regressions in the regression suite and these need to be addressed. Either they are acceptable changes (you've introduced a recursion guard to address infinite recursion) in which case you should run the regression tests as described above and then commit the changes to the output files. Otherwise they are true regressions and you will have to figure out what changes caused them and fix them.

I've only cursorlily looked at your changes but will try to get back to them this evening. A general comment: when adding performance improvements of this nature, it is very helpful to add comments to explain why things are being done in performant but ways, rather than the more "natural" but less performant ways. Consider adding comments to your changes.

Finally, if you have a list of places where you think `SortedSet` or `SortedList` are being used inappropriately, please do let me know about them. I suggest creating a new issue in the Reko issue tracker and adding them there.

Again, thanks for the PR and your interest in Reko. Please don't be shy to add any comments you may have. I can be reached here, at https://gitter.im/uxmal/reko, or my email which is present in the Git logs.

VladRassokhin:
Hi John, 
Thank you for fast response.

Maybe creating PR was not best idea cause it's like just bunch of changes which spotlights some performance related problems and proposed fixes and should be improved further.

Talking about cache it's just copy of another cache you have somewhere nearby, AFAIR in DataTypeComparer.

About change around MapChanged: one of event listeners recreated code model on each newly processed asm instruction, that's AFAIU waste of time.

Next time will share thoughts on gitter first.

uxmal:
Noted, Vlad. I hate to be a pain in the behind, but do you think you could separate commit f1ce561154a5a7ee157cf5ac7a673c3cf0523d2b into a separate PR? As you mentioned, that change improves performance considerably by reducing the rebuilding of the text model. As for the other changes, I will take a peek at the data type inference regressions and see if they are "OK" or if we have to abandon the changes.

VladRassokhin:
@uxmal I've created separate PR #565

PR #560, opened 2018-01-29
Read relocation data only if the segment flags indicate they exist.

dmakarov:


src/ImageLoaders/MzExe/NeImageLoader.cs
```diff
@@ -582,6 +582,8 @@ bool LoadSegment(NeSegment seg, MemoryArea mem, SegmentMap imageMap)
                 mem,
                 seg.Address.Selector.Value.ToString("X4"),
                 access);
+            if ((seg.Flags & 0x100) == 0)
+                return true;
             var rdr = new LeImageReader(
```
uxmal: Could you add a symbolic name for the "magic" 0x100 constant here?

uxmal:
Hello Dmitri. Thanks for the PR! I added a small request for a symbolic constant -- I'm sure that will be easy for you to add.

dmakarov:
updated.

uxmal:
Looks good. Thanks for the contribution!

PR #561, opened 2018-01-29
Improvements around Borland SymbolLoader

VladRassokhin:
* Failure in loading Borland debug symbols should not fail 'open file' action
* Improve function type parsing - actually cause of exception

### DIFF
--- a/src/ImageLoaders/MzExe/MsdosImageLoader.cs
+++ b/src/ImageLoaders/MzExe/MsdosImageLoader.cs
@@ -135,7 +135,14 @@ namespace Reko.ImageLoaders.MzExe
                 ep.NoDecompile = true;
             }
 
-            LoadDebugSymbols(results.Symbols, addrLoad);
+			try
+			{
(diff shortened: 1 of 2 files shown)


uxmal:
Thanks for chasing this down and submitting the PR. I will polish the changes a little once the merge is done.

src/ImageLoaders/MzExe/MsdosImageLoader.cs
```diff
@@ -135,7 +135,14 @@ public override RelocationResults Relocate(Program program, Address addrLoad)
...
 
-            LoadDebugSymbols(results.Symbols, addrLoad);
+			try
+			{
+				LoadDebugSymbols(results.Symbols, addrLoad);
+			}
+			catch (Exception e)
+			{
+				Console.WriteLine("Failed to load Borland debug symbols: {0}", e.Message);
+			}
```
ptomin: These are tabs. Should be indented by spaces

uxmal:
I made a change in ef4013a5d683c that, as a side effect, removed the tabs.

PR #562, opened 2018-01-30
MipsRelocator: skip relocation if .dynamic is not present

smx-smx:


uxmal:
Short and to the point! Thanks for the PR!

PR #565, opened 2018-02-03
Speedup 'Mark as Procedure' action by not recreating text model on each change

VladRassokhin:


ptomin:
Maybe it would be better to improve recreation of `NestedTextModel` in `CombinedCodeViewer` rather than `ImageMap` modification
@uxmal What do you think about it?

src/Core/ImageMap.cs
```diff
@@ -251,6 +251,29 @@ public bool TryFindItemExact(Address addr, out ImageMapItem item)
 		{
             return Items.TryGetValue(addr, out item);
 		}
+	    
+	    
+        private bool _mapChangedEventHandlerPaused;
+	    private bool _mapChangedPendingEvents;
```
ptomin: Names of fields should not be started with underscore

src/Core/ImageMap.cs
```diff
@@ -251,6 +251,29 @@ public bool TryFindItemExact(Address addr, out ImageMapItem item)
...
+	    {
+	        _mapChangedEventHandlerPaused = true;
+	    }
+
+	    public void UnpauseEventHandler()
+	    {
+	        _mapChangedEventHandlerPaused = false;
+	        if (_mapChangedPendingEvents) MapChanged.Fire(this);
+	        _mapChangedPendingEvents = false;
+	    }
```
ptomin: This is tab, should be indented by spaces

ptomin:
And also `Cmd_MarkProcedureTests.Cmasp_Do` test is failed

src/Core/ImageMap.cs
VladRassokhin: Is there code style definition available somewhere?
AFAIU spaces should be used instead of tabs, right?

uxmal: Yes, we prefer spaces to tabs. However, there is no coding style guide yet. As we attract more contributors, it becomes necessary to write one. I will open an issue to track this.

uxmal:
@VladRassokhin's PR highlights a few issues that probably need addressing. In no particular order:

The `NestedTextModel` is expensive to create because we are re-building the model for the whole `Program` when perhaps only small parts of it actually need updating. @VladRassokhin's approach is to defer the recreation until the Reko scanning engine is done. The way he's implemented makes sense since currently the `Mark as Procedure` command is synchronous on the UI thread (regrettably blocking it -- see below). Deferring the re-creation of the model until the synchronous scanning operation is complete certainly speeds it up by avoiding work that wouldn't be visible in UI until the `Mark as Procedure` command is completed anyway.

A more advanced approach would attempt to locate the piece of the text model that actually is affected by each notification and only recompute that and its parents (recursively). This would require some effort to implement but would help address the cost issue. There still a non-zero cost to rebuild parts of the textmodel, but has the advantage that the changes in the text model could be displayed "live" as changes take place. However, the current synchronous nature of `Mark as Procedure` prevents the UI thread from refreshing, so the results still wouldn't be visible until the scanning is completed.

Unlike the "primary" scanning phase, which happens in a background thread, the `Mark as Procedure` command is executed synchronously on the UI thread. This is unfelicitous, as it is quite likely that it might take a lot of time and thereby block the UI until scanning is complete. There is also no way to cancel the scanning, the way you can cancel the "primary" scanning phase. 

Today, the "primary" scanning phase pops up a modal box, displaying what is going on and providing the user with a "Cancel" button. However, because it is modal, the user can't browse around while the scanner is working. On a large binary, this could be quite irritating.  I've opened #567 to track this and related multithreading concerns.

uxmal:
@VladRassokhin : I've finally pulled the PR -- apologies for not getting around to it sooner. I fixed the broken unit test and added a TODO reflecting my comments above. Thanks for pointing out this performance issue.

VladRassokhin:
@uxmal Thank you!

PR #572, opened 2018-02-08
Improve Borland debug symbols loading

VladRassokhin:


src/ImageLoaders/MzExe/Borland/SymbolLoader.cs
```diff
@@ -589,18 +589,71 @@ private void LoadTypeTable(LeImageReader rdr, string[] names)
...
+                case TID_PARRAY:
+                    {
+                        ++i;
+                        var filler = rdr.ReadByte();
+                        var elementType = rdr.ReadLeUInt16();
+                        var arrayIndexType = rdr.ReadLeUInt16();
+                        var f1 = rdr.ReadLeUInt16();
+                        var f2 = rdr.ReadLeUInt16();
+                        var f3 = rdr.ReadLeUInt16();
+                        SerializedType indexDataType;
```
uxmal: Are f1, f2, and f3 filler? Do you think you could label them "filler1", "filler2", and "filler3" to make it clearer? Alternatively, just throw the values away.

VladRassokhin: Sure, renamed them

uxmal:
Thanks for the PR, @VladRassokhin !

PR #578, opened 2018-03-03
Make scanner work on unit test with common jump

ptomin:
- Create scanner test with common jump 
- Always emit `goto` statements when block scanning to avoid error during block splitting


ptomin:
Merged. John, please review it. If you have any remarks, I'll create new PR to fix them.

PR #580, opened 2018-03-04
Reload project browser after finish of decompilation

ptomin:
So all decompiled procedures will be in the tree in the same way as after scanning


uxmal:
Looks good.

PR #581, opened 2018-03-06
Revert regression tests

ptomin:
Fix the problem described in https://github.com/uxmal/reko/commit/fe0bdea0fc29374288f4c3d4cfd2986f54944f1b#commitcomment-27917155
That commit caused losing of ALL subjects changes done in master branch


ptomin:
@uxmal I think it could be merged

PR #583, opened 2018-03-09
Define FPU stack variable after hell node

ptomin:
Define FPU stack variable after hell node (e.g. indirect calls)
It could improve output data for `VCExeSample`


uxmal:
Looks good.

PR #587, opened 2018-03-18
Add FPU stack pointer to list of trashed registers on win32

ptomin:
Add `FPU` stack pointer to list of trashed registers on `win32`


ptomin:
@uxmal I'm not good at `Win16` and `Msdos` platform. Should `Top` register be added to trashed registers on these platforms?

uxmal:
Indeed, the `Top` register is modifiable on those platforms too. For instance, a function that returns a `long double` would return it on the FPU stack on both Microsoft and Borland compilers of that era.

ptomin:
> Please add support for this in the MsdosPlatform and Win16Platform classes.

Done

PR #589, opened 2018-03-25
MacOS Classic - expand compressed Application global data into A5 World

gbody:
updating 25 march 2018


ptomin:
There is no useful commits. It seems to be sent by mistake

PR #590, opened 2018-03-25
Rename local variables of AdjustRegisterAfterCall

ptomin:
`sp` -> `reg`
`stack pointer` -> `register`
This method can adjust any register.


uxmal:
Looks fine. I may be needing this in `analysis-development` soon.

PR #593, opened 2018-03-29
Rearrange methods of CoalescerTests: move RunTest to begin

ptomin:
Rearrange methods of `CoalescerTests` in the same order as in `analysis-development` branch


ptomin:
Good. 
P. S. @uxmal What about `Float80Tests.F80_NegInfinity`?

ptomin:
> P. S. @uxmal What about Float80Tests.F80_NegInfinity?

Fixed in 422220a

PR #594, opened 2018-03-29
Use invariant culture for string representation of float80 in unit tests

ptomin:
Use invariant culture for string representation of `float80` in unit tests to avoid failure on `Travis`


uxmal:
Thanks for hunting this one down, @ptomin.

PR #591, opened 2018-03-26
expand compressed Application global data into A5 World

gbody:

PR #596, opened 2018-03-30
updated A5World expand

gbody:
I think I have things setup better on my end now.


src/Environments/MacOS/ResourceFork.cs
```diff
@@ -342,7 +342,151 @@ public void Dump()
...
                 platform.A5Offset = jt.BelowA5Size;
                 segmentMap.AddSegment(a5world);
+
+                // Search segmentMaps for %A5Init CODE Segment and unpack global data to A5World
+                foreach ( var a5dataSegment  in segmentMap.Segments.Values)
+                {
+                    if (a5dataSegment.Name.Length >= 12 )
+                    {
+                        if (a5dataSegment.Name.Substring(5,7) == "%A5Init")
+                        {
```
ptomin: What about https://github.com/uxmal/reko/pull/591#discussion_r177093255?

gbody: I am scanning the segments looking for the A5%Init code segment, not sure of a better way.
I did remove the embedded expansion DO loop into a seperate procedure.

gbody: I just noticed in testing that I left a wrong value for bank size, it should be 0x00010000. So this needs to be updated. The testing was to check as I tried to display an output message if there was a problem with the information found in the A5%Init code segment. Not sure of the right way yet.

ptomin: Would not be better to write

```
if (a5dataSegment.Name.Length < 12 )
    continue;
if (a5dataSegment.Name.Substring(5,7) != "%A5Init")
   continue;
if (<somethig-else>)
  continue;
...
<A5Init found>
```
to avoid nested `if`s

src/Environments/MacOS/ResourceFork.cs
```diff
@@ -343,28 +343,29 @@ public void Dump()
...
+                                a5dr.Seek(a5dheaderOffset - 2);
                                 var a5dbelow = a5dr.ReadBeUInt32();
                                 var a5dbankSize = a5dr.ReadBeUInt32();
                                 var a5doffset = a5dr.ReadBeUInt32();
                                 var a5dreloc = a5dr.ReadBeUInt32();
                                 var a5dhdrend = a5dr.ReadBeUInt32();
-                                if (a5dbankSize == 0x00010001)
+                                if (a5dbankSize == 0x00010000)
                                 {
                                     a5Expand(a5dr, a5dbelow);
```
ptomin: Should be `A5Expand` according to style guide

src/Environments/MacOS/ResourceFork.cs
ptomin: The other possible solution is split it to separate methods. Something like
```
var seg = findA5InitSegment(<args>);
if (seg != null)
    readA5InitSegment(seg, <other args>);
```

src/Environments/MacOS/ResourceFork.cs
```diff
@@ -347,50 +347,42 @@ public void Dump()
...
-                            {
-                                // starting code bytes incorrect or 
-                                // invalid offset to compressed global data - outside of memory segment.
-                                // no global compressed data 
-                            }
-                        }
+                        // Starting code bytes incorrect or 
+                        // Invalid offset to compressed global data - outside of A5World memory segment.
+                        // No global compressed data 
+                        break;
```
ptomin: Are you sure that should be `break` rather than `continue`? There was not loop break at previous version of your code.

src/Environments/MacOS/ResourceFork.cs
```diff
@@ -347,50 +347,42 @@ public void Dump()
                 // If CODE segment found, unpack global data to A5World
                 foreach ( var a5dataSegment  in segmentMap.Segments.Values)
                 {
-                    if (a5dataSegment.Name.Length >= 12 )
+                    if (a5dataSegment.Name.Length < 12)
+                        continue;
+                    if (a5dataSegment.Name.Substring(5, 7) != "%A5Init")
+                        continue;
```
ptomin: Nice! What about next refactoring stage: methods extracting. Something like
```
var seg = findA5InitSegment(<args>);
if (seg != null)
    readA5InitSegment(seg, <other args>);
```

uxmal:
@gbody : I refactored your code a little for readability and committed. Thanks for the PR! How would you like to be credited in the `AUTHORS` file in the root of the Reko repository?

gbody:
John,
not need for any credit.

Regards
Geoff Body
  ----- Original Message -----

PR #597, opened 2018-03-31
Improve unit test for SsaMutators

ptomin:
Improve unit test for `SsaMutators`: use `SsaProcedureBuilder` to avoid `SsaTransform` call

### DIFF
--- a/src/UnitTests/Analysis/SsaMutatorTests.cs
+++ b/src/UnitTests/Analysis/SsaMutatorTests.cs
@@ -22,6 +22,7 @@ using NUnit.Framework;
 using Reko.Analysis;
 using Reko.Core;
 using Reko.Core.Code;
+using Reko.Core.Expressions;
 using Reko.UnitTests.Mocks;
 using Rhino.Mocks;
 using System;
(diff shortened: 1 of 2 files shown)


src/UnitTests/Analysis/SsaMutatorTests.cs
```diff
@@ -37,111 +39,67 @@ namespace Reko.UnitTests.Analysis
...
-		defs: a1_6,a3_7
-	r63_5 = r63_1
-	return
-	// succ:  ProcedureBuilder_exit
-ProcedureBuilder_exit:
+            var sExp = @"
+sp_1 = fp
+call a (retsize: 4;)
+	uses: a,sp_1
+sp_5 = sp_1
```
ptomin: @uxmal Looks better, is not it?

uxmal: I still want to see the ssa state of at least the sp identifier to make sure the ssa state has been updated correctly

ptomin: `CheckUses` method verifiers ssa state. So we'are sure the ssa state has been updated correctly

ptomin: @uxmal I have replaced `CheckUses`  with `Check` which verifies both uses and definitions. So now is no more reason to print `ssa` state

uxmal: Can we change the word `Check` to `Validate` as that is more idiomatic in situations like this where you are validating the "sanity" of the code state. I have no objections to the changes otherwise.

ptomin: OK

ptomin: Done in e9958a30

uxmal: Thanks!

PR #598, opened 2018-03-31
Move common code to Reko.UnitTests.Mocks

ptomin:
Move common code to `Reko.UnitTests.Mocks`


ptomin:
Merged. Any objections, John?

PR #599, opened 2018-04-01
Check ssa definitions

ptomin:
- Add check `ssa` definitions method to `SsaState`
- Add check `SsaState` method which checks both uses and definitions 
- Use `Check` rather than `CheckUses` (except `ValuePropagatorTests`)


ptomin:
Merged. Any remarks, John?

PR #602, opened 2018-04-03
Validate ssa definitions when running Value Propagation Tests

ptomin:
Validate `ssa` definitions when running Value Propagation Tests(`ValidateUses` -> `Validate`)


uxmal:
Unit tests are failing. Are you looking into that?

ptomin:
... and all tests are passed

PR #606, opened 2018-04-08
Load comments from project files

ptomin:
- Create unit test for `Project Loader` with user comments


ptomin:
Merged. Any remarks, John?

PR #612, opened 2018-04-16
Fix broken MixedCodeDataModel tests with comments

ptomin:


ptomin:
Merged. Any objections, John?

PR #613, opened 2018-04-16
Create MixedCodeDataModel Position class and used it instead of address

ptomin:
It contains both address and comment offset


uxmal:
Looks good!

PR #615, opened 2018-04-19
Make CodeFormatter work with multiline comments

ptomin:


src/Core/StringEx.cs
```diff
@@ -0,0 +1,44 @@
...
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Reko.Core
+{
```
uxmal: Consider inlining this in the class that is using it. Only if we find that we are splitting into lines in many places does it make sense to create an extension class for `System.String`.

ptomin: I'm going to use it at `CodeFormatter` also

uxmal: I'm still unhappy about tacking this onto the `string` class. Consider using it inlining it in the two places you're using it.

ptomin: OK

uxmal:
One unit test failing currently.

ptomin:
It's OK. I have created it to reproduce error

ptomin:
Ready to merge?

PR #619, opened 2018-04-21
Create hotkey for comment editing

ptomin:
Use `Ctrl+T` (T - Text) as hotkey for comment editing. Do you agree, John?


uxmal:
Well, `Ctrl+T` is already used to mark memory regions of memory with a **T**ype in the memory control. I was actually going to suggest the simple command `;` (semicolon). What do you think?

ptomin:
`;` without `Ctrl`?

uxmal:
Yes, a simple semicolon. This is not a general text editor, so commands without a `Ctrl` prefix is fine.

ptomin:
`OemSemicolon` works on my machine

uxmal:
OK, let's go with that for now. It's possible that German or French users may complain. I'm not going to worry until we get bug reports for it. Having the capability of changing the key bindings will certainly help.

PR #622, opened 2018-04-21
Misc changes

smx-smx:
Fixed typo in x86 opcode name
Temporarly commented out mips relocator (broken) and suffix array generation (slow, and currently unused)


uxmal:
Thanks for the PR!

PR #624, opened 2018-04-30
Modify scanner to trash registers after call

ptomin:
- Create block work item unit test to verify trashed registers after call


subjects/regressions/reko-90/PP.c
```diff
@@ -15021,7 +15020,7 @@ selector fn0800_A4F6(word16 bp, selector ds, word16 wArg02, word16 wArg04, word1
 word16 fn0800_A53C(byte al, selector ds, segptr32 ptrArg02, byte bArg06, word16 wArg08)
 {
 	word16 ax_24;
-	msdos_unknown_2143();
+	__syscall(33);
```
ptomin: This is regression. Is not it, John?

uxmal: Yes. We should still be generating `msdos_unknown_2143` here.

ptomin: Fixed in 85dbdeb

uxmal:
Looks good.

PR #628, opened 2018-05-02
Do not allow external procedure signature to be null

ptomin:
Do not allow external procedure signature to be `null`. It caused `NullReferenceException` when executing version from `analysis-development` branch


ptomin:
Type variables were not created for `null` signatures
Now some `null` signatures become not 'null' signatures with invalid
parameters. So there is equivalence classes renumbering in regression
tests
Hope it's OK, John

PR #629, opened 2018-05-05
Remove null check for procedure signature

ptomin:
Remove null check for procedure signature. Procedure signatures could not be null.


ptomin:
There is equivalence classes renumbering in regression
tests. Hope it's OK

uxmal:
Equivalence classes tend to shift when you make changes like you did. This looks fine.

PR #634, opened 2018-05-13
Modify CFG Cleaner to do not eliminate critical instructions

ptomin:
- Create unit test to reproduce removing procedure call by CFG Cleaner
- Move `CriticalInstruction` to Reko.Core.Code
- Modify CFG Cleaner to do not eliminate critical instructions


uxmal:
Nice catch Pavel. I'm surprised this hadn't been discovered earlier.

PR #636, opened 2018-05-19
Check if there are dead statements uses

ptomin:
Check if there are uses which is not in the procedure statements list
during ssa state uses validation


ptomin:
There are errors in `CondtionCodeEliminator` unit tests (`shl_rcl/shr_rcr`). New check discovers dead statements in the uses list. Great!
The fix is coming...

ptomin:
Fixed. Please review it, John.

uxmal:
Nice catch Pavel. It's going to become more and more important to keep the SSA state consistent, and these validation functions you're contributing are a great help.

PR #637, opened 2018-05-25
Appveyor: switched configuration to x64

smx-smx:


ptomin:
@smx-smx  John has fixed #644 issue. So `Appveyor` can be switched now to `x64`.
Closing this.

ptomin:
See #668

PR #639, opened 2018-05-27
Replace FPU definitions after trivial phi removing

ptomin:
- Create unit test to reproduce FPU definitions corruption after trivial phi removing
- Replace FPU definitions after trivial phi removing


uxmal:
Thanks for fixing this oversight.  The new unit test should catch future slip ups

PR #640, opened 2018-06-01
 Make MemoryAccess and SegmentedAccess immutable 

ptomin:
Backwalker causes several pointers  to same expressions like `Mem[ebp + 8]`. `SsaTransform` in `analysis-development` failed because of unexpected instruction change when other instruction is changing. Making `MemoryAccess` immutable could prevent such errors


uxmal:
Makes sense to move toward immutable data structures. Nice initiative

PR #641, opened 2018-06-01
Create new instances of MemoryAccess rather than modifying existing ones [analysis-development]

ptomin:
- Create new instances of `MemoryAccess` rather than modifying existing ones
- Remove unused methods from `ValuePropagationTests `


uxmal:
Thanks for this Pavel

Issue #643, opened 2018-06-06
Constant.Create does not support bit-sized constant.


chostelet:
Only 1, 2, 4, 8, 16, 32 byte-wide constants are currently allowed.

PR #645, opened 2018-06-06
Bugfixes on 538

chostelet:
Made correction on arch-option in PIC18 examples.
Defined a default PIC name in case none is specified.
Removed spurious NUnit NuGet packages
Corrected PrimitiveType.CreateWord() but kept CreateWordFromBits() method until issue #643 closed.


src/Gui/Windows/Forms/DecompilerMenus.cs
```diff
@@ -0,0 +1,535 @@
+#region License
+/*
```
ptomin: This file should not be under source control

chostelet: Exact. It's back ever and ever….. I've removed it.

ptomin:
Minor note. Why вщ you push so many unnecessary commits like `Merge remote-tracking branch 'JohnK/master'`? Actually there are only 3 useful commits (since 54a3969). It would be better (but it is not obligatorily of course) to clean history

ptomin:
Now there is no error with unrecognised input parameter `arch_option`. Thanks!
But there are still `System.IO.FileNotFoundException`s when loading `PIC` binaries.

chostelet:
Pavel,
Concerning the multiple commits, this was unexpected. However you have to understand that I'm totally not fluent with Git (that's an understatement). With the instrumental help of John, my history has already been cleanup. So forgive me if my commits are not "state-of-the-art", I simply hope they are not too confusing.

chostelet:
I'm correcting the Common.Items (typo on MicrochipPIC.dll) however we should still have an issue with .dcproject files loading (had a discussion with John yesterday - kind-of chicken-egg situation he's looking at).

chostelet:
Confirmed. We're getting an exception (ArgumentNullException) which is "expected" as-of-today per the discussion I had with John. Loading .dcproject file for PIC is not working yet.

ptomin:
Now `.dcproject` file is successfully loaded on `Travis` but failed on `Appveyor`:
```
error: An error occurred during decompilation. System.TypeLoadException: 
Could not load type 'Reko.Arch.MicrochipPIC.Common.PICArchitecture' from 
assembly 'Reko.Arch.MicrochipPIC, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

chostelet:
The only issue I could see was the fact that the compiler for MicrochipPIC libs was reset to default….

I've not found a way to force VS to stick to a specific compiler version - globally on a solution.

chostelet:
That was an attempt to make both AppVeyor and Travis-CI "happy" with those decompilation result's files.
Reverting back.

chostelet:
And sorry for the bad noise.

chostelet:
I'm totally confused with the way AppVeyor is saying it miss the result files (PIC18Extd.asm, etc) and Travis-CI being happy not having them.  Why this difference ?

chostelet:
I've no clue why Appveyor is not satisfied with those untracked files (log lines 5383 thru 5397)

ptomin:
>Travis-CI being happy not having them. Why this difference ?

Reko failed to load PIC files on `Travis`
```
error: An error occurred during decompilation. System.TypeLoadException:
Could not load type 'Reko.Arch.MicrochipPIC.Common.PICArchitecture' from 
assembly 'Reko.Arch.MicrochipPIC, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

ptomin:
And it loads successfully on `Appveyor`

ptomin:
> I've no clue why Appveyor is not satisfied with those untracked files (log lines 5383 thru 5397)

The decompilation result files should be committed

chostelet:
I'm totally lost.
This is the same library (source) for every environment. All the 6582 tests complete successfully on my machine. And Travis-CI says - at the end - it is happy; even if with a TypeLoadException. This exception does not occur either on my machine, nor in Appveyor (which is still not happy for obscure reasons).

Is there any way to "debug" under Travis-CI ?

chostelet:
Concerning the decompilation result files, I'm perceiving a difference (discrepancy) between Travis-CI and Appveyor. One seems not to expect those files to be under git, while the other suggests to add them. ?!?

ptomin:
> This is the same library (source) for every environment.

It's not the same. `Travis` is linux, `Appveyor` is windows

ptomin:
> 'm perceiving a difference (discrepancy) between Travis-CI and Appveyor. One seems not to expect those files to be under git, while the other suggests to add them. ?!?

It failed on `Travis` and so output files were not produced. It work successfully on `Appveyor` and so output files were created. So there is the difference.
When `Travis` failure will be fixed, the same outputs will be produced both on `Appveyor` and `Travis`

chostelet:
Pavel, I agree that the object (lib) is not the same, but the sources are the same (same class name, same namespace).
Now, I'm discovering on my machine that depending on the configuration (Debug, Release, TravisRelease) the compiler is not always set to C#7.0 as it was agreed. Don't know if that is the cause. Should try.

ptomin:
C#7 couldn't cause such issues

chostelet:
Again, do you know of a way to debug under Travis CI ? 
I can't figure out the cause the TypeLoadException

ptomin:
If `C#7` is not set then it use default language version which is `C#7` for VS 2017

chostelet:
It is C# 7.3 on my VS.

chostelet:
I think I found my mistake. Had no reference to MicrochipPIC library for CmdLine driver… :-(
I'm ashame.

chostelet:
TypeLoadException solved…. but that's a never-ending story.
How to get ride of those files as they are modified each time decompile is ran? That's normal they are modified. So why Travis-CI and Appveyor are unhappy?
Should I remove those subjects folders? I'm about to give-up.

uxmal:
We can look at this together this evening if you like.

ptomin:
Finally it passed all tests!

chostelet:
Yes, hard learning.
The final issue was that I was using my examples PIC18 hex files with WindowsDecompiler which generate different results as what the call to 'decompile.exe' in 'subject.cmd' is producing. In the future, I'll first take note not to use those hex files folders in interactive mode (but some copy of them, not under Git).

chostelet:
However, I've opened some issues that I found during those many attempts...

chostelet:
This push is to resynchronize with latest 'reko/uxmal/master'

uxmal:
@chostelet: tests are passing now -- is this PR ready for integration to `master`?

chostelet:
Should be OK. However, I couldn't remove the temporary CreateWordFromBits() method (in PrimitiveType.cs) because as of now Constant.cs does not support bit-sized values (e.g. Constant24 like word24 for a 24-bit wide item). And modifying Constant.cs, is probably not something I would do by myself.

uxmal:
You've opened an issue for that, so it's OK. I'll go ahead and merge.

PR #649, opened 2018-06-11
Print pointer bit size to make unit tests discover more bugs

ptomin:
Print pointer bit size to make unit tests discover more bugs. 12 unit tests become failed. These are `varargs` parsers. The reason of errors is 5c12b9a


src/Environments/Windows/MsPrintfFormatParser.cs
```diff
@@ -47,7 +47,9 @@ protected override DataType MakeDataType(PrintfSize size, char cDomain)
         {
             if (cDomain == 'S')
             {
-                return program.TypeFactory.CreatePointer(PrimitiveType.WChar, base.pointerSize);
+                return program.TypeFactory.CreatePointer(
```
uxmal: Consider making CreatePointer bit-sized as well -- we should be in general be moving towards bit orientation when working with primitive types. `Pointer` is basically a `PrimitiveType` of domain `Pointer` with extra type information. Then we don't have to do all these divisions.

src/Environments/Windows/MsWprintfFormatParser.cs
```diff
@@ -51,8 +51,12 @@ protected override DataType MakeDataType(PrintfSize size, char cDomain)
             {
             case 'c': return PrimitiveType.WChar;
             case 'C': return PrimitiveType.Char;
-            case 's': return program.TypeFactory.CreatePointer(PrimitiveType.WChar, base.pointerSize);
-            case 'S': return program.TypeFactory.CreatePointer(PrimitiveType.Char, base.pointerSize);
+            case 's': return program.TypeFactory.CreatePointer(
+                PrimitiveType.WChar,
```
uxmal: Same comment here; make `TypeFactory.CreatePointer` accept a bit size rather than a byte size.

src/Libraries/Libc/ScanfFormatParser.cs
```diff
@@ -114,7 +114,7 @@ protected virtual DataType MakeDataType(PrintfSize size, char cDomain)
             case 's':
                 return program.TypeFactory.CreatePointer(
                     size == PrintfSize.Long ? PrimitiveType.WChar : PrimitiveType.Char,
-                    pointerSize);
+                    pointerSize / DataType.BitsPerByte);
```
uxmal: ...and now you can remove this division.

src/Libraries/Libc/ScanfFormatParser.cs
```diff
@@ -162,7 +162,7 @@ protected virtual DataType MakeDataType(PrintfSize size, char cDomain)
             }
             return program.TypeFactory.CreatePointer(
                 PrimitiveType.Create(domain, bitSize),
-                pointerSize);
+                pointerSize / DataType.BitsPerByte);
```
uxmal: ...and same comment here. No need for division.

src/Libraries/Python/PyArgFormatParser.cs
```diff
@@ -109,7 +109,7 @@ public class PyArgFormatParser : IVarargsFormatParser
 
         private DataType Ptr(DataType pointee)
         {
-            return new Pointer(pointee, pointerSize);
+            return new Pointer(pointee, pointerSize / DataType.BitsPerByte);
```
uxmal: And same comment here.

ptomin:
I have changed `TypeFactory.CreatePointer` in 9ebde38. Should same change be applied to `Pointer`

ptomin:
But there are too many uses of `Pointer`. Lets left it for future development

uxmal:
Yes, that can be done as a separate step. Otherwise, vse khorosho?

ptomin:
>Otherwise, vse khorosho?

Просто отлично!

PR #650, opened 2018-06-11
Make Pointer class bit-sized

ptomin:
Finally I have made `Pointer` class bit-sized


uxmal:
See, that wasn't that bad, was it? :) Thanks for the PR, Pavel.

PR #659, opened 2018-06-23
Fix build error on Mac

### LINKED ISSUE #658: Build errors on Visual Studio for Mac
I encountered 2 build errors while trying to build on macOS High Sierra with Visual Studio for Mac:
1. cmake cannot find capstone which is already installed (via homebrew). Adding the line `set(PKG_CONFIG_EXECUTABLE /usr/local/bin/pkg-config)` before the `pkg_check_modules` command in CMakeLists.txt works around this.
2. Later in the build process (AfterBuild?), `libArmNative.so` is not found. However, `libArmNative.dylib` is present at its place. I worked around this by replacing `libArmNative.so` in `Drivers/Common.items`.

With both problems resolved, the program compiles successfully, but crashes immediately upon launch. (WindowsDecompiler Debug|x64)

EDIT: Fix formatting.

wildptr:
Fix #658

### DIFF
--- a/src/Drivers/Common.items
+++ b/src/Drivers/Common.items
@@ -1,10 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!--
-This file contains the definitions of all files required for a 
+This file contains the definitions of all files required for a
 complete Reko distribution. The ItemGroups defined here are consumed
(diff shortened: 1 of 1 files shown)


uxmal:
Thanks for the PR!

PR #660, opened 2018-06-29
Eliminates usage of CreateWordFromBits method 

chostelet:
Eliminates temporary method to get bit-size constants and binary words as needed specifically by PIC processor (registers).


uxmal:
Thanks for taking the time to clean up the PR, Christian.

PR #666, opened 2018-08-05
Remove obsolete FPU methods

ptomin:
- Represent FPU accesses explicitly when `FXTRACT` rewriting instead of using obsolete method `GrowFpuStack`
- Remove obsolete FPU methods


uxmal:
Nice job getting rid of the obsolete code. I guess this is one way of telling me "get the `analysis-develoment` branch merged into `master` :). Once I'm back in Stockholm. Honest!

PR #672, opened 2018-08-18
Do not use MemoryStorage for global variables

ptomin:
- Create SSA transform unit test with global variables
- Do not use `MemoryStorage` for global variables


uxmal:
Again I'm away over the weekend. I will look at this on Monday. There are some issues that need to be addressed before we can accept the PR.

ptomin:
I think, this PR could be merged and then you can used it as start of new changes (if needed)

uxmal:
Yes, I think we make this a starting point..

PR #675, opened 2018-08-26
Pass storage as separate argument of UsedRegisterFinder.Classify

ptomin:
- Pass `SsaState` rather than `SsaTransform` to `UnusedOutValuesRemover`
It makes unit testing easier.
- Add `UnusedOutValuesRemover` test with call which definition storage differs from identifier storage
- Pass storage as separate argument of `UsedRegisterFinder.Classify`


uxmal:
Looks good.

PR #687, opened 2018-09-16
SsaEvaluationContext: do not search imported procedures in FPU stack

ptomin:
4 unit tests failed during trying to resolve FPU stack access as import reference on X86 real mode. Modification of `SsaEvaluationContext` to avoid search imported procedures in FPU stack could fix that


src/Decompiler/Analysis/SsaEvaluationContext.cs
```diff
@@ -74,6 +74,9 @@ public Expression GetValue(MemoryAccess access, SegmentMap segmentMap)
         {
             if (access.EffectiveAddress is Constant c)
             {
+                //Do not search imported procedures in FPU stack
+                if (access.MemoryId.IsFpuStack)
+                    return access;
```
uxmal: To avoid "infecting" the code with x86 specifics, consider testing the domain of the `MemoryId.Storage` and only continue  if it is `StorageDomain.Memory`. The domain of the `ST` identifier is not that value.

src/Core/Expressions/Identifier.cs
```diff
@@ -159,6 +159,8 @@ public MemoryIdentifier(string name, DataType dt, Storage stg) : base(name, dt,
         {
         }
 
+        public bool IsFpuStack => Name.StartsWith("ST"); //$HACK: gross hack but we have to start somewhere.
+
```
uxmal: I wish to avoid this X86 specific code in this General purpose class. See my other comment.

src/Decompiler/Analysis/SsaEvaluationContext.cs
ptomin: Fixed in cced050

PR #690, opened 2018-09-18
Fixed: crash on select AmigaOS props page

wdigger:


uxmal:
Thanks for the bugfix. Would you like to be credited in the AUTHORS file? If so, would you like your name in cyrillic (Юрий Клушин, I think?)

wdigger:
@uxmal thanks, but it's not necessary ;)

PR #692, opened 2018-09-22
CallApplicationBuilder: return fallback for stack argumens not found in the map

ptomin:
Argument can be not found in the map due to bugs at previous stages.
Create `stack{offset} = <Invalid>` statement in this case.
Return `stack{offset}` as fallback
This is better than exception throwing


uxmal:
@ptomin : could you please add a unit test for this code?

ptomin:
Yes, of course!

ptomin:
Done in 0601d24

PR #693, opened 2018-09-28
CallApplicationBuilder: return fallback for register arguments not found in the map

ptomin:
- Create unit test for `CallRewriter` to verify not found register argument

### DIFF
--- a/src/Decompiler/Analysis/CallApplicationBuilder.cs
+++ b/src/Decompiler/Analysis/CallApplicationBuilder.cs
@@ -41,6 +41,7 @@ namespace Reko.Analysis
         private readonly Dictionary<StorageDomain, CallBinding> defs;
         private readonly Dictionary<StorageDomain, CallBinding> uses;
         private Dictionary<StorageDomain, CallBinding> map;
+        private bool bindUses;
 
         public CallApplicationBuilder(SsaState ssaCaller, Statement stmCall, CallInstruction call, Expression callee) : base(call.CallSite, callee)
         {
@@ -59,18 +60,30 @@ namespace Reko.Analysis
 
         public override Expression Bind(Identifier id)
         {
-            return With(uses, id.Storage);
+            return WithUses(id.Storage);
         }
 
         public override OutArgument BindOutArg(Identifier id)
         {
-            var exp = With(defs, id.Storage);
+            var exp = WithDefinitions(id.Storage);
             return new OutArgument(arch.FramePointerType, exp);
         }
 
         public override Expression BindReturnValue(Identifier id)
         {
-            return With(defs, id.Storage);
+            return WithDefinitions(id.Storage);
+        }
+
+        private Expression WithUses(Storage stg)
+        {
+            this.bindUses = true;
+            return With(uses, stg);
+        }
+
+        private Expression WithDefinitions(Storage stg)
+        {
+            this.bindUses = false;
+            return With(defs, stg);
         }
 
         private Expression With(Dictionary<StorageDomain, CallBinding> map, Storage stg)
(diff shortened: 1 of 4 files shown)


ptomin:
- Return fallback for register arguments not found in the map in the same way as in #692 
    Create `{registerName} = <Invalid>` statement.
    Return `{registerName}` as fallback

src/UnitTests/Analysis/CallRewriterTests.cs
```diff
@@ -783,6 +783,47 @@ public void CrwStackArgumentNotFound()
...
+
+            When_RewriteCalls(ssa);
+
+            var sExp =
+            #region Expected
+@"main_entry:
+body:
+	// Failed to bind call argument.
+	// Please report this issue at https://github.com/uxmal/reko
+	arg = <invalid>
```
uxmal: As explained elsewhere, instead of emitting `arg = <invalid>` here, emit a `def arg` in `main_entry`.

ptomin:
Done in e0dc20a

src/Decompiler/Analysis/CallApplicationBuilder.cs
```diff
@@ -59,18 +60,30 @@ public CallApplicationBuilder(SsaState ssaCaller, Statement stmCall, CallInstruc
...
+            return WithDefinitions(id.Storage);
+        }
+
+        private Expression WithUses(Storage stg)
+        {
+            this.bindUses = true;
+            return With(uses, stg);
+        }
+
+        private Expression WithDefinitions(Storage stg)
```
uxmal: Could you add a comment why we don't want to bind uses in this case?

src/Decompiler/Analysis/CallApplicationBuilder.cs
```diff
@@ -198,5 +213,24 @@ private Statement InsertStatementBefore(Instruction instr, Statement stmBefore)
...
+        {
+            if (ssaCaller.Identifiers.TryGetValue(id, out var sid))
+                return sid;
+            sid = ssaCaller.Identifiers.Add(id, null, null, false);
+            var def = new DefInstruction(sid.Identifier);
+            var block = ssaCaller.Procedure.EntryBlock;
+            sid.DefStatement = block.Statements.Add(0, def);
+            return sid;
+        }
     }
```
uxmal: Lines 229-233 do the same as lines in the method `NewDefInstruction`1460-1464 in `SsaTransform.cs`. I suggest refactoring both, and creating a new method on `SsaState` called `MakeDefInstruction` which both can call.

src/Decompiler/Analysis/CallApplicationBuilder.cs
ptomin: >why we don't want to bind uses in this case?

Because this this is return, so we should bind it with `defs`, is not it? Do you want me to add such comment before `WithDefinitions` method?

src/Decompiler/Analysis/CallApplicationBuilder.cs
ptomin: I have moved `EnsureSsaIdentifier` method from `SsaTransform` to `SsaState` in 3ffada84
The name `EnsureSsaIdentifier` is suggested by you (see https://github.com/uxmal/reko/pull/667#discussion_r209508554)
So I do not rename it to `MakeDefInstruction `

ptomin:
- I have do requested refactoring of `EnsureSsaIdentifier`.
- I do not properly understand your request about code commenting

uxmal:
Thanks for this Pavel. Don't worry about the comment. I will review on my end.

PR #695, opened 2018-09-30
Avoid InvalidOperationException throwing if return/out argument storage was not found

ptomin:
- Create unit test for `CallRewriter.RewriteReturns` with not found return.
It throws following exception
```
   System.InvalidOperationException : Sequence contains no elements
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at Reko.Analysis.CallRewriter.RewriteReturns(SsaState ssa)
```
- CallRewriter.RewriteReturns: avoid `InvalidOperationException` throwing 
if return/out argument storage was not found. Return `Constant.Invalid` instead

### DIFF
--- a/src/Decompiler/Analysis/CallRewriter.cs
+++ b/src/Decompiler/Analysis/CallRewriter.cs
@@ -345,11 +345,13 @@ namespace Reko.Analysis
                 if (idRet != null && !(idRet.DataType is VoidType))
                 {
                     var idStg = reachingBlock.Value
-                        .Where(cb => cb.Storage.Covers(idRet.Storage)).First();
+                        .Where(cb => cb.Storage.Covers(idRet.Storage))
+                        .FirstOrDefault();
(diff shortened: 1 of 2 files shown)


uxmal:
I'm going to accept the PR since I agree that it's better than just giving up and dying. I'm also investigating why these are happening in the subject suite. I'm looking at the Z-80 CB80.com sample now, which exhibits a lot of problems related to this PR.

ptomin:
> I'm looking at the Z-80 CB80.com sample now, which exhibits a lot of problems related to this PR.

Strange. It cannot exhibits a lot of problems related to this PR on my machine. `CB80.com` failed at early stage with `NotImplementedException` on `GetSubFlags` methos

uxmal:
I have fixed the `GetSubFlags` issue on my machine after which I run into problems where we're incorrectly deducing live out registers.

PR #696, opened 2018-10-03
ExpressionEmitter: modify IAdd to create right operand as word constant

ptomin:
This is the same modification as for `ISub` in #338
Difference between `ISub` and `IAdd` caused incorrect simplification of `FPU`
stack pointer in `analysis-development`. `TOP + 1` expression differs from
`TOP + 0x01` and so `PHI(TOP + 1, TOP + 0x01)` expression could be
simplified

### DIFF
--- a/src/Core/Expressions/ExpressionEmitter.cs
+++ b/src/Core/Expressions/ExpressionEmitter.cs
@@ -67,7 +67,7 @@ namespace Reko.Core.Expressions
         /// <returns>A binary expression for the sum.</returns>
         public BinaryExpression IAdd(Expression left, int right)
         {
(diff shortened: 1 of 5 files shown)


ptomin:
John, I have seen you comment about it https://github.com/uxmal/reko/commit/50044b5b3cb39e4f6e0ece26a7b6d426264afad5#r30747047
I'll try to fix it

ptomin:
@uxmal I have modified code so that there is no chages in Unit test outputs. Is it enough?

uxmal:
This looks OK. There are a lot of places where `IAddS` can be used.

PR #699, opened 2018-10-06
Make MapLinearAddressToAddress work for address out of segments (e.g. 0)

ptomin:
Make `MapLinearAddressToAddress` work for address out of segments (e.g. 0)
Avoid throwing of `ArgumentOutOfRangeException`


src/Core/SegmentMap.cs
```diff
@@ -170,17 +170,17 @@ public bool IsExecutableAddress(Address addr)
...
-                    long offset = (long)linearAddress - (long)seg.Address.ToLinear();
+                    offset = (long)linearAddress - (long)seg.Address.ToLinear();
                     return seg.Address + offset;
                 }
             }
-            throw new ArgumentOutOfRangeException(
-                string.Format("Linear address {0:X8} is not in known segment.",
-                linearAddress));
+            offset = (long)linearAddress - (long)BaseAddress.ToLinear();
+            return BaseAddress + offset;
```
uxmal: This is not going to work well on x86 real-mode binaries. However, let's consider not changing ```Statement.LinearAddress``` to ```Statement.Address``` and changing the type to ```Address```. We need to review code to see if there is any need for linear addresses at all (apart from the memory savings of course)

uxmal: Also, it might be better to fix the cause of all the linearAddresses being 0. I believe it is because phi assignments aren't being given a real linear address. It should be the linear address of the basic block in which they are located.

ptomin: I'll try to avoid linear addresses to be `zero`. See #700

ptomin:
I have reverted my fix for `MapLinearAddressToAddress`. But tests are still useful. Could be merged John?

PR #704, opened 2018-10-11
 Pass block address instead of procedure entry address to statement 

ptomin:
- Set address of entry and exit blocks in `Procedure` constructor 
- Pass block address instead of procedure entry address to statement. Address of each block should be always set
- Set address of intra statement block during scanning

### DIFF
--- a/src/Decompiler/Analysis/LiveCopyInserter.cs
+++ b/src/Decompiler/Analysis/LiveCopyInserter.cs
@@ -1,6 +1,6 @@
 #region License
 /* 
- * Copyright (C) 1999-2018 John Källén.
+ * Copyright (C) 1999-2018 John KÃ¤llÃ©n.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


src/Decompiler/Analysis/LiveCopyInserter.cs
```diff
@@ -56,7 +56,7 @@ public int IndexOfInsertedCopy(Block b)
 		public Identifier InsertAssignmentNewId(Identifier idOld, Block b, int i)
 		{
             var stm = new Statement(
-                (b.Address ?? b.Procedure.EntryAddress).ToLinear(),
+                b.Address.ToLinear(),
                 null,
```
uxmal: Much better!

src/Decompiler/Scanning/BlockWorkitem.cs
```diff
@@ -1083,11 +1083,8 @@ private Block FallthroughBlock(Address addrSrc, Procedure proc, Address fallthru
         private Block AddIntraStatementBlock(Procedure proc)
         {
             var label = ric.Address.GenerateName("l", string.Format("_{0}", ++extraLabels));
-            var fallthru = new Block(proc, label)
-            {
-                IsSynthesized = true
-            };
-            proc.ControlGraph.Blocks.Add(fallthru);
+            var fallthru = proc.AddBlock(ric.Address, label);
+            fallthru.IsSynthesized = true;
```
uxmal: Let's add an `AddSyntheticBlock` to `Procedure`.

ptomin: Done in 3e9e27c

PR #705, opened 2018-10-13
Modify parameter reading: use SLICE when extracting smaller bitvectors

ptomin:


uxmal:
Молодец! This is a good improvement

PR #707, opened 2018-10-19
Ensure that the result of trivial phi simplification is alive     

ptomin:
- Create `SsaTransform` test with loop with break and return. It failed to simplify trivial phi functions
- Ensure that the result of trivial phi simplification is alive
This result could be removed during recursive trivial phi simplification
Simplify phi function until identifier is not in the list of
removed phi identifiers

### DIFF
--- a/src/Decompiler/Analysis/SsaTransform.cs
+++ b/src/Decompiler/Analysis/SsaTransform.cs
@@ -1357,21 +1357,9 @@ namespace Reko.Analysis
             /// <returns>true if the phi function was trivial</returns>
             private bool TryRemoveTrivial(SsaIdentifier phi, out SsaIdentifier sid)
             {
-                Identifier same = null;
-                var phiFunc = ((PhiAssignment)phi.DefStatement.Instruction).Src;
+                var phiFunc = ((PhiAssignment) phi.DefStatement.Instruction).Src;
                 DebugEx.PrintIf(trace.TraceVerbose, "  Checking {0} for triviality", phiFunc);
-                foreach (Identifier op in phiFunc.Arguments)
-                {
-                    if (op == same || op == phi.Identifier)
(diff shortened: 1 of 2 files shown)


src/Decompiler/Analysis/SsaTransform.cs
```diff
@@ -1357,29 +1357,19 @@ private static void GeneratePhiFunction(SsaIdentifier phi, SsaIdentifier[] sids)
...
-                        continue;
-                    if (same != null)
-                    {
-                        sid = null;
-                        return false;
-                    }
-                    same = op;
-                }
-                if (same == null)
+                if (!phiFunc.Arguments.Where(a => a != phi.Identifier).Any())
```
uxmal:
```suggestion
                if (!phiFunc.Arguments.Any(a => a != phi.Identifier))
```
may be slightly more performant.

src/Decompiler/Analysis/SsaTransform.cs
```diff
@@ -1357,29 +1357,19 @@ private static void GeneratePhiFunction(SsaIdentifier phi, SsaIdentifier[] sids)
...
-                        continue;
-                    if (same != null)
-                    {
-                        sid = null;
-                        return false;
-                    }
-                    same = op;
-                }
-                if (same == null)
+                if (!phiFunc.Arguments.Any(a => a != phi.Identifier))
```
uxmal: The double negative is confusing here, now. Isn't this the same as saying 
```suggestion
                if (phiFunc.Arguments.All(a => a == phi.Identifier))
```

ptomin: Changed. See 9f76399

ptomin:
@uxmal I have done requested changes

PR #709, opened 2018-11-06
Disable false positive Armadillo v1.71 PEiD signature

wesinator:
The Armadillo v1.71 packer signature from PEiD is a false positive, it detects MSVC compiled binaries.


uxmal:
It may well be that `IMAGE_FILE_MACHINE_I386.xml` may be out of date in other ways. Thanks for the PR!

wesinator:
Yea, if I find anything else I'll try to make a PR. The Armadillo v1.71 is a big false positive pretty much everywhere that uses PEiD signatures.

PR #711, opened 2018-11-07
Multiple architectures

uxmal:
This PR implements initial support for multiple processor architectures in the same binary. This opens the way to supporting MacOSX "fat binaries" with PPC + x86, ELF AArch32 binaries with a mixture of T32 and A32 instructions, MSDOS DPMI programs with 16-bit real mode bootstrappers and 32-bit application code (among others).

A central notion is that we try to avoid using `Program.Architecture` as much as possible, relying instead on whatever architecture Reko has discovered a particular blob has. 

There is certainly more work to do. For instance, there should be a way for a user to add an architecture to a list of known architectures from the user interface. When more than 1 architecture is is known, the Reko gui should ask the user "which of the N processor architectures" when executing the "Mark Procedure" command. I'm expecting we'll discover a lot of these issues as the feature starts being used.

I'm requesting a code review from the reviewers to make sure I'm not missing anything obvious and to get you acquainted with the code. I encourage you to ask questions if anything is unclear.


src/ImageLoaders/Elf/ElfLoader.cs
```diff
@@ -878,7 +882,7 @@ public override void Dump(TextWriter writer)
         public override List<string> GetDependencyList(byte[] rawImage)
         {
             return Dependencies;
-        }
+            }
```
ptomin: Incorrect indentation

src/ImageLoaders/Elf/ElfLoader.cs
```diff
@@ -718,7 +722,7 @@ protected string rwx(uint flags)
             sb.Append((flags & 2) != 0 ? 'w' : '-');
             sb.Append((flags & 1) != 0 ? 'x' : '-');
             return sb.ToString();
-        }
+    }
```
ptomin: Incorrect indentation

src/ImageLoaders/Elf/ElfLoader.cs
```diff
@@ -1148,13 +1152,13 @@ public override ElfSymbol LoadSymbol(ulong offsetSymtab, ulong symbolIndex, ulon
...
-                        Value = sym.st_value,
-                        Size = sym.st_size,
-                    });
+                {
+                    Name = RemoveModuleSuffix(ReadAsciiString(stringtableSection.FileOffset + sym.st_name)),
+                    Type = (ElfSymbolType)(sym.st_info & 0xF),
+                    SectionIndex = sym.st_shndx,
+                    Value = sym.st_value,
+                    Size = sym.st_size,
+                });
```
ptomin: Incorrect indentation

src/ImageLoaders/Elf/ElfLoader.cs
```diff
@@ -504,22 +511,18 @@ public void ConstructGotEntries(Program program, SortedList<Address, ImageSymbol
                         symbols[addrGot] = gotSym;
                         Debug.Print("{0}+{1:X4}: Found GOT entry {2}, referring to symbol at {3}", 
                             gotStart, addrGot-gotStart, gotSym, symbol);
-                        program.ImportReferences.Add(addrGot, new NamedImportReference(addrGot, null, symbol.Name));
+                            program.ImportReferences.Add(addrGot, new NamedImportReference(addrGot, null, symbol.Name));
```
ptomin: Incorrect indentation

src/Core/Program.cs
```diff
@@ -226,6 +227,15 @@ public TypeLibraryDeserializer CreateTypeLibraryDeserializer()
...
         }
 
+
+        /// The processor architectures that exist in the Program. 
+        /// <remarks>
+        /// Normally there is only one architecture. But there are examples
+        /// of binaries that have two or more processor architectures. E.g.
+        /// "fat binaries" on MacOS, or ELF binaries with both ARM32 and 
+        /// Thumb instructions.</remarks>
+        public Dictionary<string, IProcessorArchitecture> Architectures { get; }
```
ptomin: Unused property?

uxmal: This will be used shortly: users will be able to add new architectures if they want.

PR #714, opened 2018-11-10
Replace only overlapping stack definitions to improve performance

ptomin:
Code profiling shows that the most time is taken by stack definitions
`IntervalTree` enumeration in `ReplaceBy` method

### DIFF
--- a/src/Decompiler/Analysis/SsaTransform.cs
+++ b/src/Decompiler/Analysis/SsaTransform.cs
@@ -1492,19 +1492,34 @@ namespace Reko.Analysis
                             alias.SsaId = idNew;
                         }
                     }
-                    foreach (var de in bs.currentStackDef.ToList())
-                    {
-                        if (de.Value.SsaId == sidOld)
-                        {
-                            de.Value.SsaId = idNew;
-                        }
-                    }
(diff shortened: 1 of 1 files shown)


src/Decompiler/Analysis/SsaTransform.cs
```diff
@@ -1508,6 +1502,28 @@ private void ReplaceBy(SsaIdentifier sidOld, SsaIdentifier idNew)
...
+            {
+                if (!(sidOld.Identifier.Storage is StackStorage stack))
+                    return;
+                var offsetInterval = Interval.Create(
+                    stack.StackOffset,
+                    stack.StackOffset + sidOld.Identifier.DataType.Size);
+                var ints = bs.currentStackDef
+                    .GetIntervalsOverlappingWith(offsetInterval)
+                    .ToArray();
+                foreach (var de in ints)
```
uxmal: If this is performance critical, perhaps we shouldn't materialize the `ints` into an array, but feed the intervals directly into the following `foreach` loop.

ptomin: Done. See 4eba33e

PR #719, opened 2018-11-25
Modify CallRewriter to add uses of out parameter store to Ssa State

ptomin:
Modify `CallRewriter` to add uses of out parameter store to `Ssa State`. This is the fix for https://github.com/uxmal/reko/blob/56a7d8b304bd13c6dabdc0f30528561af2566804/subjects/regression.log#L57


uxmal:
Pavel, in commit c39f2094 I resolved this in a way I think is more correct. In particular, the `out` parmeters shouldn't create `def` nodes in the entry block of a procedure as this would imply that they are live on entry to the function, while in fact they are live out.

PR #720, opened 2018-11-28
Make Cast immutable

ptomin:
Make `Cast` immutable. It should cause correct work of one more regression test on `analysis-development` branch. SsaTransform in `analysis-development` failed because of unexpected instruction change when other instruction is changing.


uxmal:
Nice work. I think it is a good idea to make more code immutable, to promote thread safety.

PR #723, opened 2018-12-13
Analysis phi arguments

uxmal:
The arguments to the SSA PhiFunction are strongly associated with the basic block from which the arguments "arrive". However, this was not expressed clearly in code. The new PhiArgument class makes the origin of each PhiArgument explicit.


uxmal:
This also mirrors the way LLVM explicitly mentions, for each incoming argument of an SSA phi-function, the basic block from whence it came.

uxmal:
Thanks for looking at this. I'll abandon the PR and push against `master`, then merge into `analysis-development`.

Issue #735, opened 2019-02-25
SsaIdentifierTransformer produces incorrect results on a MIPS sample


uxmal:
The following IR code, which is paraphrasing a MIPS binary I'm investigating, is being incorrectly rewritten by `SsaIdentifierTransformer.Transform`:

```
    r2 = Mem0[0x0220200:word32]    // fetch some data
    r4 = r2                      // the compiler sucks, or optimizer turned off
    r2 = Mem0[0x00400000:word32]  // our function pointer is also in memory
    r25 = r2    // as mentioned, the compiler sucks
    call r25    // adhere to the MIPS ABI
```

After SSA and value propagation this becomes:
```
    r2_1 = Mem0[0x0220200:word32]  
    r4_2 = r2_1
    r2_3 = Mem3[0x00400000:word32]
    r25_4 = r2_3
    call r25_4 uses:r2_1
```
Then, Reko discovers the signature for the  procedure whose pointer is at `0x00400000` and makes it available in the import resolver. Value propagation is applied again, and we expect the following:
```
    r2_1 = Mem0[0x0220200:word32]  
    r4_2 = r2_1
    r2_3 = my_function
    r25_4 = my_function
    my_function(r2_1)
```
but we instead get:
```
    r2_1 = Mem0[0x0220200:word32]  
    r4_2 = r2_1
    r2_3 = my_function
    r25_4 = my_function
    my_function(my_function)
```
I've a unit test in `ValuePropagationTests.cs` that captures this behaviour. @ptomin: could you take a look?

ptomin:
Interesting case! I'll look at it

ptomin:
John, it's very strange decision to use `SsaIdentifierTransformer` after `CallApplicationBuilder`. It was designed for `IndirectCallRewriter` to convert expressions like `<fn>(ecx, Mem[esp+4]` to `<fn>(ecx_1, Mem[esp_2+4]`. 
There is already correct SSA before `VP` running. So we need the same adjustment as
https://github.com/uxmal/reko/blob/a1fc088a97285833a394bcc95ce8dc37424f7c34/src/Decompiler/Analysis/CallRewriter.cs#L307
I'll try to fix it (maybe tomorrow)

ptomin:
> John, it's very strange decision to use `SsaIdentifierTransformer` after `CallApplicationBuilder`

[...]

PR #737, opened 2019-02-26
Fix building of application in "ValuePropagator"

ptomin:
Fixes for #735


ptomin:
I think, it could be merged. There is `StackOverflow` during executing of `PDP-11/space.dcproject` but it's another issue. And it was before I have made this changes

PR #739, opened 2019-03-03
ValuePropagator: adjust stack after call rewriting

ptomin:
`ValuePropagator` should emit `esp = esp + delta` statement after call rewriting


uxmal:
Thanks for this, Pavel. Could you merge into `analysis-development` as well?

ptomin:
Yes, of course. I'm going to do it this evening

ptomin:
Done. See c9b42ef

Issue #73, opened 2015-11-13
Discussion: output file organization


uxmal:
Right now, if you feed reko an executable `foo.exe`, it dumps the following files in the same directory:
* `foo.dcproject` - a Reko project file.
* `foo.c` - all the executable code reko was able to discover
* `foo.asm` - assembly language for all instructions that were "reached" by the scanner
* `foo.dis` - dump of the IL Language; this will eventually go away.
* `foo.h` - all data types that have been inferred.
* `foo.globals.c` - all global data types 

Some of the issues with this are:
* It is very cluttering to scatter all these files into the directory the executable lives 
in. Suggested improvement: by default, create a subdirectory `reko` under the current directory, 
and place all output files there. Naturally, the location of the files can be user overridden.
* The file extension for assembler files is hardwired to `.asm`. Each `Platform` should be able
to override this to indicate its preferred file extension.
* MS-DOS and Windows16 binaries contain the infamous segments. However, in this case 
the segments could serve as a guide in how to decompose the binary into smaller pieces. Rather than
a big `foo.c` file, why not a `foo.0800.c` (or `foo.seg0800.c`?) containing the code of 
the code segment loaded at address 0800:0000. Similarly, the code segment at 0F3A:0000 
would be written to `foo.0F3A.c`; naturally, global variables in segment 1DA3:0000 would
go into `foo.1DA3.c`.
* The same applies for more modern file formats like ELF and PE, where the sections are 
clearly labelled with a name. Thus, a typical ELF binary would result in `foo.text.c`, `foo.data.c`,
and `foo.bss.c` files being created.
* PE image files contain resources. Today, Reko knows how to parse them, but  doesn't 
save the contents of the resources to disk. It probably should create a `foo.rc` file and then 
create a `resources` subdirectory with all the file resources.

To recap, the output of a MS-DOS binary would look like this:
```
.
+- foo.exe
+- foo.dcproject
+- reko
   +- foo.0800.c
   +- foo.0F3A.c
   +- foo.1DA3.c
   +- foo.asm
   +- foo.h
```
while an ELF binary appears like this:
```
.
+- foo
+- foo.dcproject
+- reko
  +- foo.bss.c
  +- foo.data.c
  +- foo.h
  +- foo.s
  +- foo.text.c
 ```
and a PE binary appears like this:
```
.
+- foo.exe
+- foo.dcproject
+- reko
  +- foo.asm
  +- foo.bss.c
  +- foo.data.c
  +- foo.h
[...]

PR #750, opened 2019-03-17
Introduce OutputDirectory concept to Reko projects

Cairn23:


ptomin:
Build error:
```
Gui/Forms/FakePhasePageInteractors.cs(105,71): error CS0535: 'FakeInitialPageInteractor' does not implement interface member 'InitialPageInteractor.OpenBinary(string, string)' [/home/travis/build/uxmal/reko/src/UnitTests/UnitTests.csproj]
Gui/Forms/FakePhasePageInteractors.cs(105,71): error CS0535: 'FakeInitialPageInteractor' does not implement interface member 'InitialPageInteractor.OpenBinaryAs(string, string, LoadDetails)' [/home/travis/build/uxmal/reko/src/UnitTests/UnitTests.csproj]
Gui/Forms/FakePhasePageInteractors.cs(105,71): error CS0535: 'FakeInitialPageInteractor' does not implement interface member 'InitialPageInteractor.Assemble(string, string, Assembler)' [/home/travis/build/uxmal/reko/src/UnitTests/UnitTests.csproj]
```

src/Decompiler/Decompiler.cs
```diff
@@ -269,20 +270,24 @@ public void Assemble(string fileName, Assembler asm)
...
+        protected Project CreateDefaultProject(string fileNameWithPath, string outputDir, Program program)
         {
-            program.Filename = fileName;
-            program.EnsureFilenames(fileName);
+            program.Filename = fileNameWithPath;
+
+            FileInfo fif = new FileInfo(fileNameWithPath);
+            string fileName = fif.Name;
+
+            program.EnsureFilenames(outputDir + "\\" + fileName);
```
ptomin: Platform-specific code

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -182,5 +192,78 @@ private bool NeedsScanning()
...
         }
+
+        private void ProcessResources(ProgramResourceGroup prg, string outputDir)
+        {
+            try
+            {
+                if (prg.Name == "PE resources")
+                {
+                    // Create the dir
+                    Directory.CreateDirectory(outputDir + "\\resources");
```
ptomin: Platform-specific code

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -182,5 +192,78 @@ private bool NeedsScanning()
...
+                            }
+                            break;
+                            case "FONT":
+                            {
+                                WriteResourceFile(outputDir + "\\resources", "Font", ".bin", pr);
+                            }
+                            break;
+                            case "NEWBITMAP":
+                            {
+                                WriteResourceFile(outputDir + "\\resources", "NewBitmap", ".bmp", pr);
```
ptomin: Platform-specific code

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -182,5 +192,78 @@ private bool NeedsScanning()
...
+            }
+            catch (Exception)
+            {
+
+            }
+        }
+
+        private void WriteResourceFile(string outputDir, string ResourceType, string ext, ProgramResourceGroup pr)
+        {
+            Directory.CreateDirectory(outputDir + "\\" + ResourceType);
```
ptomin: Platform-specific code

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -182,5 +192,78 @@ private bool NeedsScanning()
...
+            foreach (ProgramResourceGroup pr1 in pr.Resources)
+            {
+                foreach (ProgramResourceInstance pr2 in pr1.Resources)
+                {
+                    // write the bytes to file
+                    try
+                    {
+                        using (var fs = new FileStream(outputDir + "\\" + ResourceType + "\\" + pr1.Name + ext, FileMode.Create, FileAccess.Write))
```
ptomin: Platform-specific code

src/Gui/MruList.cs
```diff
@@ -31,16 +31,22 @@ namespace Reko.Gui
 	/// </summary>
 	public class MruList
 	{
-		private int itemsMax;
-		private List<string> items;
+        public struct fileOutputDir
+        {
+            public string filename;
+            public string outputDir;
+        };
```
ptomin: I think, changing `Mru` format is not good idea. It's just the list of last opened files and should contain file names only

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -529,6 +533,12 @@ public string ProjectFileName
             set { projectFileName = value; }
         }
 
+        public string ProjectOutputDir
+        {
+            get { return projectOutputDir; }
+            set { projectOutputDir = value; }
+        }
```
ptomin: `ProjectOutputDir` (and `projectOutputDir`) looks useless. I can't see any reading of this property (except line 702)

src/Decompiler/Decompiler.cs
```diff
@@ -269,20 +270,24 @@ public void Assemble(string fileName, Assembler asm)
...
             byte[] image = loader.LoadImageBytes(fileName, 0);
             var program = loader.LoadRawImage(fileName, image, null, raw);
-            Project = CreateDefaultProject(fileName, program);
+            Project = CreateDefaultProject(fileName, outputDir, program);
             eventListener.ShowStatus("Raw bytes loaded.");
             return program;
         }
 
-        protected Project CreateDefaultProject(string fileName, Program program)
+        protected Project CreateDefaultProject(string fileNameWithPath, string outputDir, Program program)
```
ptomin: Maybe it's not best solution to add `output directory` parameter to `CreateDefaultProject`. We should create _default_ project in this method. So we need to auto-genetate some path (and maybe it should be done in `Program.EnsureFilenames`)

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -106,35 +107,44 @@ public override bool LeavePage()
...
             {
-                isOldProject = Decompiler.Load(file);
+                isOldProject = Decompiler.Load(file, outputDir);
             });
             if (Decompiler.Project == null)
+            {
                 return false;
+            }
+
+            ProgramResourceGroup prg = Decompiler.Project.Programs[0].Resources;
```
ptomin: Should be `foreach` programs, is not it?

src/Decompiler/Decompiler.cs
uxmal: I suggest you use `Path.Combine` here to avoid the platform specific code.

src/Gui/Forms/InitialPageInteractor.cs
uxmal: Consider using `Path.Combine` to build the path name. 

In Reko, to make program code unit testable, we don't write code directly to `System.IO.Directory` or  `System.IO.File`. Instead, we call methods on the interface `IFileSystemService`. This way we can simulate file systems in unit tests. At the top of this method, you should add the line:
```C#
var fsSvc = Services.RequireService<IFileSystemService>();
```
and then replace this call with:
```C#
fsSvc.CreateDirectory(Path.Combine(outputDir, "resources");
```
Since that method doesn't exists on `Reko.Core.Services.IFileSystemService` yet, please add it to the interface and the class `Reko.Core.Services.FileSystemImpl`.

It is possible the user may want to rename the `resources` directory, but we can look at that later.

src/Gui/Forms/InitialPageInteractor.cs
uxmal: I like the initiative of extracting the resources of the binary as well. We will have to look into what would be needed to do the same for MacOS classic files.

This is PE-specific code, which should probably be handled by code in `PEImageLoader`. Since a proper abstraction is not yet in place for the manipulation embedded resources, let's not worry about it now. Please add a `//$TODO: move into PEImageLoader` comment so this doesn't get forgotten.

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -29,14 +29,15 @@
...
 {
     public interface InitialPageInteractor : IPhasePageInteractor
     {
-        bool OpenBinary(string file);
-        bool OpenBinaryAs(string file, LoadDetails details);
-        bool Assemble(string file, Assembler asm);
+        bool OpenBinary(string file, string outputDir);
+        bool OpenBinaryAs(string file, string outputDir, LoadDetails details);
+        bool Assemble(string file, string outputDir, Assembler asm);
     }
```
uxmal: Same comment as before: the `OutputDirectory` should be a property internal to the `Project`.

src/Gui/Forms/InitialPageInteractor.cs
uxmal: Because this method may have opened a project file, you should iterate across all the `Program`s in the `Decompiler.Project`.

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -182,5 +192,78 @@ private bool NeedsScanning()
...
+                            break;
+
+                            default:
+                                break;
+                        }
+                    }
+                }
+            }
+            catch (Exception)
+            {
```
uxmal: If the user runs out of file space or there is a permission error, they should be notified. Instead of discarding the exception, use:
```C#
catch (Exception ex) {
    var uiSvc = Services.RequireService<IDecompilerShellUiService>();
    uiSvc.ShowError(ex"A resource file could not be written.");
}
```

src/Gui/Forms/InitialPageInteractor.cs
uxmal: Again, avoid using `FileStream` directly. Instead, use `IFileSystemService`:
```C#
var fsSvc = Services.RequireService<IFileSystemService>();
var filename = Path.Combine(outputDir, ResourceType, pr1.Name + ext);
using (var fs = fsSvc.CreateFileStream(filename, FileMode.Create, FileAccess.Write))
...
```

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -182,5 +192,78 @@ private bool NeedsScanning()
...
+                    try
+                    {
+                        using (var fs = new FileStream(outputDir + "\\" + ResourceType + "\\" + pr1.Name + ext, FileMode.Create, FileAccess.Write))
+                        {
+                            fs.Write(pr2.Bytes, 0, pr2.Bytes.Length);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+
```
uxmal: Again, the exception is being dropped on the floor. Since this method is being called in a loop, a better reporting mechanism is to use the `IDiagnosticsService.Error` method:
```
catch (Exception ex) {
    var diagSvc = Services.RequireService<IDiagnosticsService>();
    diagSvc.Error(ex, "An error occurred while displaying the program.");
}
```
Consider returning a `bool` from this method: return `true` if write was successful, and `false` if it failed. The calling method should `and` together all returned values from this method, and if the total result is `false`, show a dialog box with the message "One or more resource files was not saved successfully."

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -219,26 +220,25 @@ public IMainForm MainForm
...
         {
             try
             {
                 CloseProject();
                 SwitchInteractor(InitialPageInteractor);
-                if (openAction(file))
-                {
-                    ProjectFileName = file;
-                }
+                openAction(file);
```
uxmal: If a file could not be opened we should not be setting the name of the opened file property.

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -286,7 +292,7 @@ public void AddMetadataFile()
             {
                 var metadata = projectLoader.LoadMetadataFile(fileName);
                 decompilerSvc.Decompiler.Project.MetadataFiles.Add(metadata);
-                RememberFilenameInMru(fileName);
+                //RememberFilenameInMru(fileName);
```
uxmal: This looks like debugging code, temporarily commenting out this line?

src/Gui/Forms/MainFormInteractor.cs
uxmal: `ProjectOutputDirectory` should be located in the `Project` class and should be used when generating the names of the output files that `Program` uses.

src/Gui/MruList.cs
uxmal: The MRU (most recently used file) list should just contain file names. The output directory name can be deduced for simple binaries, and for project files, they should contain a property called `OutputDirectoryName` which would allow users to modify the location of the directory if they want.

src/Gui/decompiler-menus.xml
```diff
@@ -74,7 +74,7 @@
   <!-- Commands ========================================================== -->
 
 	<cmd id="FileOpen" cmdSet="Reko" container="FileOpen" image="0">_Open...</cmd>
-  <cmd id="FileOpenAs" cmdSet="Reko" container="FileOpenAs">Op_en As...</cmd>
+  <cmd id="FileOpenAs" cmdSet="Reko" container="FileOpenAs">Op_en As (raw)...</cmd>
```
uxmal: "Raw" is a little misleading, as some file formats are not raw, but don't contain sufficient metadata to be loaded successfully without further hints from the user. Intel HEX files come to mind: they definitely are not raw blobs of bytes, but they don't have any indication of what processor architecture or platform to use.

uxmal:
I had to look back at issue #73 again, it's been a long time. There we discussed whether all files from a project should be dumped into a single `reko` output directory, or whether there should be a separate output directory for each binary in the project. I'm not suggesting incorporating the outcome of that discussion into this PR, but once this gets accepted and merged, we should take advantage of the momentum and just implement the features as outlined there.

ptomin:
>  I suggest the following:
> 
>     * Adding an `OutputDirectory` to the `Project` class.
> 
I think we do not need new property in `Project`. We already have
- DisassemblyFilename
- IntermediateFilename
- OutputFilename
- TypesFilename
- GlobalsFilename

We should just change them to ensure writing output files to `<Binary name>.reko` by default

uxmal:
Then we need a `ResourceFilesDirectory`, as it seems conceivable a user might want to change the output directory of recovered resources.

Cairn23:
May I suggest running the code and see how I have done things in that it creates a resource directory under the .reko, and then I have created further directory's for Bitmap icon etc

uxmal:
@Cairn23 : I think there are two concerns crossing over. The first one is the writing of PE resources: awesome stuff, I love it, let's go with it. The second is the impact of changing the API's to have an explicit output directory mentioned everywhere. @ptomin and I are saying that it could and should be done without adding that extra parameter.

src/Gui/decompiler-menus.xml
Cairn23: Whilst in the module that did the processing of  binary file and updating the input params there was a comment with TODO, so in order to clean up the code I did!

Cairn23:
I think the comment about added it to the project file is good, but the project file is not always saved, so additional work would be required to do that.

uxmal:
I've made a new branch, `pr-750`, consisting of commits 6d821f9c3a9c7064496146856e6fe56b9141b8dc
7d7f9a941f38840ee16c446cccf79b2c3503cab5 and 
 387b6c6f5e23c05c2c758379911116ddd5f6a5d7 that addresses some of the comments here. Please review those commits and test to see if it still works as expected. I was able to extract some resources on some of the files in my "zoo" of executables just fine.

Cairn23:
I have tested branch 750 and it still works for me.

uxmal:
Good, I will merge it into `master` and `analysis-development`. Thanks for the contribution!

PR #756, opened 2019-03-24
Switch to last version of mono

ptomin:
Mono bug was fixed! See https://github.com/mono/mono/issues/11663


uxmal:
Nice! Thanks for this, Pavel!

Issue #764, opened 2019-04-06
Unnecessary union in VCExeSample output


ptomin:
See https://github.com/ptomin/reko/pull/5/files#r267813480
`master` branch output:
```
c->vtbl->method04(c, c->vtbl->sum(c, a, b));
```
`analysis-development` branch output:
```
c.u0->vtbl->method04(c, c.u0->vtbl->sum(c, a, b));
```

ptomin:
Fixed in 067fd30

PR #765, opened 2019-04-06
Move UserSignatureBuilder call after creation of SSA

ptomin:
It should prevent type inference between user-defined parameters and
other expressions and creation of unnecessary unions (this is the fix for #764)


uxmal:
Nice work Pavel!

PR #768, opened 2019-04-08
Feature: use BTreeDictionary instead of SortedTree for O(n log n) complexity

uxmal:
When doing performance testing, I'm noticing that one of the "hotspots" is `SortedList.Add`. This is not surprising as we are using `SortedList` everywhere, and `SortedList` is infamously bad: it is basically [insertion sort](https://en.wikipedia.org/wiki/Insertion_sort) which degenerates to `O(n^2)` complexity when adding non-sorted data.

I spent some time writing another sorted dictionary implementation, based on [B+trees](https://en.wikipedia.org/wiki/B%2B_tree). I've only replaced a few uses of `SortedList` with `BTreeDictionary` but am already seeing 10% performance improvements when running regression tests.

Please review this to make sure there's nothing glaringly obvious missing.

### DIFF
--- a/src/Core/Lib/BTreeDictionary.cs
+++ b/src/Core/Lib/BTreeDictionary.cs
@@ -1,0 +1,750 @@
+#region License
+/* 
+ * Copyright (C) 1999-2019 John Källén.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ */
+#endregion
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+
+namespace Reko.Core.Lib
+{
+    /// <summary>
+    /// Represents a collection of key/value pairs that are sorted by the keys
+    /// and are accessible by key and by index. It's intended to be a drop-in
+    /// replacement for <see cref="System.Collections.Generic.SortedList{TKey, TValue}"/>.
+    /// </summary>
+    /// <remarks>
+    /// This class implemements most of the same API as <see cref="System.Collections.Generic.SortedList{TKey, TValue}"/>
+    /// but with much better performance. Where n random insertions into a 
+    /// SortedList have a complexity of O(n^2), the BtreeDictionary is 
+    /// organized as a B+tree, and this has a complexity of O(n log n). 
+    /// In addition, benchmark measurements show that BTreeDictionary is
+    /// about twice as fast as SortedDictionary (which also is O(n log n))
+    /// and in addition provides the IndexOf functionality from SortedList
+    /// that SortedDictionary lacks.
+    /// </remarks>
+    public class BTreeDictionary<TKey, TValue> : IDictionary<TKey, TValue>
+    {
+        private Node root;
+        private int version;
+        private int InternalNodeChildren;
+        private int LeafNodeChildren;
+        private readonly KeyCollection keyCollection;
+        private readonly ValueCollection valueCollection;
+
+        public BTreeDictionary() :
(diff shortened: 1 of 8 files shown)


src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+                }
+                return Insert(iPos + 1, rightNode.keys[0], rightNode,  tree);
+            }
+
+            public Node Add(TKey key, Node node, BTreeDictionary<TKey, TValue> tree)
+            {
+                int idx = Array.BinarySearch(keys, 1, count-1, key, tree.Comparer);
+                if (idx >= 0)
+                    throw new ArgumentException("Duplicate key.");
+                var subnode = nodes[~idx];
```
ptomin: Unused variable

src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+                int idx = Array.BinarySearch(keys, 1, count - 1, key, tree.Comparer);
+                int iPos = (idx >= 0)
+                    ? idx
+                    : (~idx) - 1;
+                var subnode = nodes[iPos];
+                var (leftNode, rightNode) = subnode.Set(key, value, tree);
+                if (rightNode == null)
+                {
+                    this.totalCount = SumNodeCounts(this.nodes, this.count);
+                    tree.Validate(this);
```
ptomin: Is it debug code?

src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+                this.keys = new TKey[children];
+                this.nodes = new Node[children];
+            }
+
+            public override (Node, Node) Add(TKey key, TValue value, BTreeDictionary<TKey, TValue> tree)
+            {
+                int idx = Array.BinarySearch(keys, 1, count-1, key, tree.Comparer);
+                int iPos;
+                if (idx >= 0)
+                    iPos = idx - 1;
```
ptomin: I think, it should be `iPos = idx` (the same as in `Set` method)

src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+                if (idx >= 0)
+                    return nodes[idx].Get(key, tree);
+                else
+                {
+                    var iPos = (~idx) - 1;
+                    return nodes[iPos].Get(key, tree);
+                }
+            }
+
+            public override (Node, Node) Set(TKey key, TValue value, BTreeDictionary<TKey, TValue> tree)
```
ptomin: `Set` is almost the same as `Add`. Maybe they are should be merged to one method with extra `bool` flag

src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+                if (idx >= 0)
+                {
+                    removed = nodes[idx].Remove(key, tree);
+                }
+                else
+                {
+                    var iPos = (~idx) - 1;
+                    removed = nodes[iPos].Remove(key, tree);
+                }
+                --this.totalCount;
```
ptomin: Maybe should be 
```
if (removed)
{
     --this.totalCount
}
```

src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+
+            public override bool Remove(TKey key, BTreeDictionary<TKey, TValue> tree)
+            {
+                int idx = Array.BinarySearch(keys, 0, count, key, tree.Comparer);
+                if (idx >= 0)
+                {
+                    --count;
+                    --totalCount;
+                    Array.Copy(keys, idx + 1, keys, idx, count - idx);
+                    Array.Copy(values, idx + 1, values, idx, count - idx);
```
ptomin: Could `idx + 1` be out of range?

src/Core/Lib/BTreeDictionary.cs
```diff
@@ -0,0 +1,750 @@
...
+                    break;
+            }
+            return ~(totalBefore + i);
+        }
+
+        public bool Remove(TKey key)
+        {
+            if (root == null)
+                return false;
+            return root.Remove(key, this);
```
ptomin: `++version`?

src/Core/Lib/BTreeDictionary.cs
uxmal: This is subtle but correct. Suppose there are two items  (`count` = 2) and we remove the item at `idx = 1`. We will decrement `count` to 1, and do `Array.Copy(keys, 1 + 1, keys, 1, 1 - 1)`. The item count of `1-1` makes the `Array.Copy` a no-op.

uxmal:
Updated the `BTree` branch with PR feedback changes. Anything else?

ptomin:
I think, it's correct now

PR #772, opened 2019-05-07
Sequence fusion

uxmal:
This commit improves Reko's output by fusing register sequences where reasonable. Fusion candidates include:
* x86 segment offset pairs: es:bx becomes es_bx
* register sequences: edx:eax
* memory accesses in adjacent stack locations: qword ptr [rsp+8], qword ptr [rsp+0x10]

The commit reduces the number of SLICE and SEQ expressions in the generated code.


subjects/regressions/reko-90/PP.c
```diff
@@ -594,30 +629,38 @@ Eq_n fn0800-0541(Eq_n ch, Eq_n di, Eq_n ds, ptr16 & bpOut, union Eq_n & dsOut)
...
+					ptrLoc06_n = SEQ(wLoc04_n, wLoc06_n);
 					break;
 				case 0x0A:
 					*((word32) ds + 10773) = 0x01;
+					ptrLoc06_n = SEQ(wLoc04_n, wLoc06_n);
 					break;
 				}
 			}
 			++*((word32) ds + 0x00002A27);
+			ptrLoc06_n = ptrLoc06_n;
```
ptomin: Redundant instruction?

uxmal:
I'm going ahead and pulling this into `analysis-development`, despite the failing CI builds. The reason is that for some reason, the type inference of `regressions\reko-90\PP.exe` is taking vastly longer. This is not caused directly by the changes in the PR. However, `PP.exe` has long had performance issues.

Recently I've been doing some performance work to speed up the scanning phase. That's now about 30% faster than in March. Looks like I will have to do similar work on the type inference phase to make TravisCI builds complete in time.

PR #774, opened 2019-05-24
Disable PEiD Armadillo packer false positive

wesinator:
https://www.zscaler.com/blogs/research/your-windows-8-packed

#709


uxmal:
Thanks for this PR. At some point, it might make sense to allow users to conveniently disable the packer signatures from the GUI and save the choices in a preferences file.

wesinator:
> Thanks for this PR. At some point, it might make sense to allow users to conveniently disable the packer signatures from the GUI and save the choices in a preferences file.

Yes - this was the last major PEiD false positive

PR #776, opened 2019-07-06
Pass pointer instead of offset to PE relocations map

ptomin:
- Relocations map keys are expected to be linear addresses.
PE relocator was passed offset instead.
- Modify `MemoryArea` to pass linear address to Relocation Dictionary instead of image offset
- Fix broken tests


uxmal:
Nice catch. That bug's been there forever.

PR #777, opened 2019-07-10
Set Conditional Transfer instruction class for FSTSW sequence

ptomin:
Set Conditional Transfer instruction class for `FSTSW` sequence.
Linear class is not correct


uxmal:
Good job. The shingle scanner will be more accurate after this fix.

PR #780, opened 2019-08-14
c2xml: fixed encoding

smx-smx:
When using standard output, the encoding might fall back to the user codepage
Force UTF8 instead


uxmal:
Thanks for this. There was an additional issue caused by a weird bug in Mono that emits text into stdout (!). I've made a workaround for that problem and now the Travis build passes.

PR #781, opened 2019-08-14
Scanner: Handle 0 clusters case

smx-smx:


uxmal:
Thanks for fixing this!

PR #783, opened 2019-08-14
Program: Consistent naming convention for the default resources direc…

smx-smx:
Use the same name as .c and .h file, thus creating a directory called
`<exeName>.resources`

This makes it possible to load multiple EXE files located in the same directory without having all resources dumped (and mixed) in a single directory called "resources"


uxmal:
This makes sense, to avoid name collisions.

PR #790, opened 2019-10-12
Simplify SLICE(SEQ(...)) even if it is in the same instruction

ptomin:
- Create unit test for simplification SLICE(SEQ(...)) in the same instruction

Expected:

`t2 = t1`

but was:

`t2 = SLICE(SEQ(t1, Mem2[0x00123400:word16]), word16, 16)`

- Simplify SLICE(SEQ(...)) even if it is in the same instruction


ptomin:
`Travis` failed on `DfaReg00005`. It should not prevent to merge this PR

ptomin:
`Travis` failure fixed in 2b8028f. Now it could be merged

ptomin:
Merged. Any objections, John?

src/Decompiler/Evaluation/SliceSequence.cs
```diff
@@ -64,9 +63,22 @@ public bool Match(Slice slice)
 
         public Expression Transform()
         {
-            ctx.RemoveIdentifierUse(idOld);
+            ctx.RemoveExpressionUse(eOld);
             ctx.UseExpression(result);
             return result;
         }
+
+        private bool IsSequence(Expression e, out MkSequence sequence)
```
uxmal: I'm thinking that possibly this could be generalized, the way `DataType.ResolveAs`, but I don't think the PR needs to be held up on that account. If this happens often enough, we should consider exposing a `ResolveAs<TExpression>` method on the `EvaluationContext` interface.

PR #791, opened 2019-10-13
Modify CallRewiter to use slice instead cast when slicing wider registers

ptomin:
- Create `CallRewriter` test with slicing wider register

Expected:

`return SLICE(ret, word16, 16)`

but was:

`return (word16) ret`

- Modify `CallRewiter` to use slice instead cast when slicing wider registers

### DIFF
--- a/src/Decompiler/Analysis/CallRewriter.cs
+++ b/src/Decompiler/Analysis/CallRewriter.cs
@@ -454,11 +454,7 @@ namespace Reko.Analysis
                     var idStg = reachingBlock.Value
                         .Where(cb => cb.Storage.Covers(idRet.Storage))
                         .FirstOrDefault();
-                    SetReturnExpression(
-                        ssa,
-                        block,
-                        idRet,
-                        idStg?.Expression ?? Constant.Invalid);
+                    SetReturnExpression(ssa, block, idRet, idStg);
                 }
                 int insertPos = block.Statements.FindIndex(s => s.Instruction is ReturnInstruction);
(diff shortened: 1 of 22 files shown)


subjects/CPM-80/CB80.c
```diff
@@ -236,7 +236,7 @@ byte fn05CE(Eq_n bc)
 		fn056B(SEQ(b_n, *globals->t166D), out bc_n);
 		globals->t166D = (word16) globals->t166D + 0x01;
 	}
-	return (byte) bc_n;
+	return SLICE(bc_n, byte, 8);
```
uxmal: This is probably an improvement, as it shows that previously, the cast was returning the `C` part of the  `BC` register pair, but here it's returning `B`, the high byte.

src/Decompiler/Analysis/CallRewriter.cs
```diff
@@ -483,16 +479,18 @@ public void RewriteReturns(SsaState ssa)
...
         {
+            var e = idStg?.Expression ?? Constant.Invalid;
             for (int i = block.Statements.Count-1; i >=0; --i)
             {
                 var stm = block.Statements[i];
                 if (stm.Instruction is ReturnInstruction ret)
                 {
                     if (idRet.DataType.BitSize < e.DataType.BitSize)
                     {
-                        e = new Cast(idRet.DataType, e);
```
uxmal: Nice catch. This was a bug before, no attention was being paid to the _offset_ of the smaller datatype.

PR #794, opened 2019-10-19
Remove implicit registers (e.g. stack pointers) from indirect call uses

ptomin:
- Remove preserved registers from external procedures with unknown signature
Do it in the same way as for indirect calls. External procedures with unknown
signature are hell nodes too.
- Remove implicit registers (e.g. stack pointers) from indirect call uses before `UsedRegisterFinder` running.
Do it in the same way as preserved registers


ptomin:
I lot of unnecessary stack assignments have gone out. John, please review and merge

src/Environments/Cpm/CpmPlatform.cs
```diff
@@ -57,7 +57,10 @@ public override MemoryMap_v1 MemoryMap
 
         public override HashSet<RegisterStorage> CreateImplicitArgumentRegisters()
         {
-            return new HashSet<RegisterStorage>();
+            return new HashSet<RegisterStorage>()
+            {
+                Architecture.StackRegister,
+            };
         }
```
uxmal: Since this class inherits from `Platform` and `Platform` now does exactly this (return the stack register as an implicit argument), you can probably remove the method altogether.

src/Environments/Windows/Win32M68kPlatform.cs
```diff
@@ -22,12 +22,13 @@ public override string DefaultCallingConvention
...
-			{
-			};
-		}
+        public override HashSet<RegisterStorage> CreateImplicitArgumentRegisters()
+        {
+            return new HashSet<RegisterStorage>
+            {
+                Architecture.StackRegister,
+            };
+        }
```
uxmal: Same comment here; `Platform`'s `CreateImplicitArgumentRegisters` already does this.

src/Environments/Cpm/CpmPlatform.cs
ptomin: There was not implementation of `CreateImplicitArgumentRegisters` in `Platform` but I have created it. See e7fd812

PR #795, opened 2019-10-21
Set size of the return address on stack to 0 on RiscV architecture

ptomin:


src/Arch/RiscV/RiscVArchitecture.cs
```diff
@@ -83,6 +83,15 @@ public RiscVArchitecture(string archId) : base(archId)
             this.StackRegister = regs[2];       // sp
         }
 
+        /// <summary>
+        /// The size of the return address if pushed on stack.
```
uxmal: Let's clarify this as "The size of the return address (in bytes)...".

ptomin: Done. See https://github.com/uxmal/reko/compare/3bf60d00810959c8dbe7d4500afd883c4cfc1690..94a748d39590c450ad7d7419807f2e05a433b52c

Issue #763, opened 2019-04-03
Indirect calls cause adding preserved registers to parameters


ptomin:
For example, we have procedure
```
...
push esi
...
call eax
...
pop esi
...
```
It is transformed during Data Flow Analysis to
```
...
dwLoc18_5 = esi
...
 // This is hell node, all stack variables was added to parameters list
 // And 'push esi' instruction is processed as stack parameter passing
call eax_13(..., dwLoc18_5, ...)
...
dwLoc18_19 = esi
...
```
and then to
```
...
 // This is hell node, all stack variables was added to parameters list
 // And 'push esi' instruction is processed as stack parameter passing
call eax_13(..., esi, ...)
...
```
So `esi` is still used and then will be added to "LiveIn" list

I suggest to remove stack arguments from indirect calls if their values are
preserved register (e.g. `ebp`, `esi`, `edi`) in similiar way as #734

uxmal:
There is a phase in `analysis-development` where we discover what registers are preserved. You can use that information to determine any memory locations that are used to preserve them. Since "hell nodes" are guesses anyway, I think it's OK to assume that indirect calls don't use such stack locations.

ptomin:
Fixed in 1a0b8da

PR #796, opened 2019-10-21
Remove stack storage of preserved registers from uses of indirect calls

ptomin:
Remove stack arguments from indirect calls if their values are
preserved register (e.g. `ebp`, `esi`, `edi`)
This is the fix for #763


uxmal:
Very nice, Pavel. The output looks much better on many subjects now.

PR #797, opened 2019-10-23
Rewrite CoaCall* tests using SsaProcedureBuilder

ptomin:


src/UnitTests/Analysis/CoalescerTests.cs
```diff
@@ -168,27 +168,43 @@ public void CoaSideEffectCalls()
...
+@"
+r4 = r2()
+call r3 (retsize: 4;)
+	uses: r2,r3,r4
+";
+            AssertProcedureCode(expected);
         }
 
         [Test]
         public void CoaCallCallee()
```
uxmal: I've been adding `[Category(Categories.UnitTests)]` to unit tests that don't require reading / writing from the file system. Now that you've gotten rid of the `*.exp` files for this test fixture, consider adding such `CategoryAttribute`

ptomin: Done in ecf88bc

PR #799, opened 2019-11-05
Remove classes which are unused in new DFA

ptomin:


uxmal:
Thanks for doing this maintenance. Once completed, you could also rename `TrashedRegisterFinder3` to `TrashedRegisterFinder`.

uxmal:
Can you look into the build failures? It's possible my earlier commits today may be causing the trouble.

PR #801, opened 2019-11-06
Remove BlockFlow and SymbolicEvaluationContext, they are unused in new analysis

ptomin:
Continue removing classes which are unused in new DFA
- Remove `BlockFlow`
- Remove obsolete properties from `ProcedureFlow`
- Remove `SymbolicEvaluationContext`


ptomin:
Merged. Any objections, John?

PR #808, opened 2019-11-21
PPC: InvalidDecoder should use CreateInvalidInstruction

smx-smx:


uxmal:
Nice job finding and fixing this!

PR #810, opened 2019-11-23
SsaTransform: search imported procedures only in Global Memory

ptomin:
Resolving FPU stack access as import reference causes `NotSupportedException` on X86 real mode.
Modification of `SsaTransform` to avoid search imported procedures in FPU stack (in the same way as #687) could fix that

### DIFF
--- a/src/Decompiler/Analysis/SsaTransform.cs
+++ b/src/Decompiler/Analysis/SsaTransform.cs
@@ -1099,7 +1099,9 @@ namespace Reko.Analysis
                 c = ea as Constant;
             }
 
-            if (c != null)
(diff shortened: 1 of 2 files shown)


uxmal:
Pavel, could you write a unit test that shows how this appears in X86 real mode?

uxmal:
Thanks for the unit test (and the PR).

ptomin:
I have done requested change. Is it time to merge, John?

uxmal:
Sorry i was waiting for the CI builds and then needed to attend to моя дорогая жена.

ptomin:
Never mind!

PR #816, opened 2019-11-30
Try to guess FPU stack delta of call instructions.

ptomin:
FPU stack delta of hell nodes is unknown. But we can guess it based on
FPU stack variable uses after call.
If we have
```
    call eax_1()
    defs: Top_3
    Extract return from FPU stack
    edx_4 = ST_2[Top_3]
    Top_5 = Top_3 + 1
```
then we can suppose that FPU stack delta is `-1`. Of course,
`ST_2[Top_3]` could be defined before indirect call, but it's unlikely

### DIFF
--- a/src/Decompiler/Analysis/FpuStackReturnGuesser.cs
+++ b/src/Decompiler/Analysis/FpuStackReturnGuesser.cs
@@ -1,0 +1,175 @@
+#region License
+/* 
+ * Copyright (C) 1999-2019 Pavel Tomin.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
+ * the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ */
+#endregion
+
+using Reko.Core;
+using Reko.Core.Code;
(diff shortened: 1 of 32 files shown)


src/Decompiler/Analysis/FpuStackReturnGuesser.cs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        public void Rewrite()
+        {
+            var fpuStack = ssa.Procedure.Architecture.FpuStackRegister;
+            if (fpuStack == null)
+                return;
+            var fpuStackIds = ssa.Identifiers
+                .Where(sid => sid.Identifier.Storage == fpuStack)
+                .Where(sid => sid.DefStatement != null)
```
uxmal: Consider fusing the two `Where` calls to a single expression:
`.Where(sid => sid.DefStatement != null && sid.Identifier.Storage == fpuStack)`

src/Decompiler/Analysis/FpuStackReturnGuesser.cs
```diff
@@ -0,0 +1,174 @@
...
+    /// <example>
+    /// If we have
+    /// <code>
+    ///     call eax_1()
+    ///         defs: Top_3
+    ///     // Extract return from FPU stack
+    ///     edx_4 = ST_2[Top_3]
+    ///     Top_5 = Top_3 + 1
+    /// </code>
+    /// then we can suppose that FPU stack delta is -1. Of course,
```
uxmal: Minor quibble: use `assume` rather than `suppose`.

ptomin: Done. See https://github.com/uxmal/reko/compare/a1d1fe1733e0c22ecb2c53b956913a801a128fb2..fb1d68656439a4435c9d99b11028745c6b00220a

src/Decompiler/Analysis/FpuStackReturnGuesser.cs
ptomin: Done. See https://github.com/uxmal/reko/compare/a1d1fe1733e0c22ecb2c53b956913a801a128fb2..fb1d68656439a4435c9d99b11028745c6b00220a

PR #817, opened 2019-12-04
Expression Simplifier: Avoid call CreateWord method if data type size is unknown

ptomin:
- Create Expression Simplifier test with cast to unknown type

Expected:

    do nothing

but was:

    throw ArgumentOutOfRangeException during CreateWord method execution

- Expression Simplifier: avoid call `CreateWord` method if data type size is unknown

### DIFF
--- a/src/Core/Types/DataType.cs
+++ b/src/Core/Types/DataType.cs
@@ -104,6 +104,15 @@ namespace Reko.Core.Types
             return dt as T;
         }
 
+        public bool IsWord()
+        {
+            if (BitSize == 0)
+                return false;
+            //$REFACTOR: CreateWord is inefficient.
+            var wordType = PrimitiveType.CreateWord(BitSize);
+            return wordType == this;
+        }
+
         protected void ThrowBadSize()
 		{
 			throw new InvalidOperationException(string.Format("Can't set size of {0}.", GetType().Name));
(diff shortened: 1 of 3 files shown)


src/Core/Types/PrimitiveType.cs
```diff
@@ -144,6 +144,14 @@ private static PrimitiveType Create(Domain dom, int bitSize, string name)
 			return shared;
 		}
 
+        public static bool IsWord(DataType dt)
```
uxmal: I'm unhappy that we have to call `CreateWord` in a predicate method (= method testing some property and returning true or false) but we don't really have a good `IsWord` type -- yet. Please add a `//$REFACTOR: CreateWord is inefficient`, make this method non-static and then move it to `DataType`:
```C#
public bool IsWord()
{
    if (this.BitSize == 0)
        return false;
    //$REFACTOR: CreateWord is inefficient. 
    var wordType = CreateWord(this.BitSize);
    return wordType == this;
}
```

ptomin: Done. See https://github.com/uxmal/reko/compare/bc60d5dbcdde7462d50bc841ac8673cf2dd3ebeb..dc646bb14758dfcf8acc216f13a49b57ad41d68a

PR #830, opened 2020-01-22
Redesigned 'genPICdb' (Unix version) + corrections

chostelet:
Implementing a Windows/Unix/NetCore compatible 'genPICdb' tool.
See ReadMe.txt file.
Still need to solve the update of the 'picdb.zip' generated file during solution buid.


uxmal:
Thanks for this @chostelet. I will continue merging work on `netStandard-ex`.

chostelet:
You should not have much problem for this other merge :-)

PR #834, opened 2020-02-05
MacOS Classic - apply relocation data to A5World

gbody:
Use the A5World relocation data to update the pointers within the A5World to point to the real addresses based on where A5World has been loaded.


src/Environments/MacOS/Classic/ResourceFork.cs
```diff
@@ -494,6 +497,79 @@ private int GetRunLengthValue(BeImageReader a5dr, ref int a5repeat)
             return runLength;
         }
 
+        private void A5Relocate(BeImageReader a5dr, UInt32 a5dbelow)
+
+        // Update pointers within A5World that point to an address within A5World, to the real/absolute address.
```
uxmal: Please move the comment to before the `private void A5Relocate(...`  and format it consistently with other comments in the code:
```C#
/// <summary>
/// Update pointers within A5 World that point to an address within A5World, to the real/absolute address. 
/// </summary>
/// <param name="a5rdr">An image reader to the segment containing the A5 World,</param>
/// <param name="a5dbelow">Offset to the location of the globals in the A5 world.</param>
```

uxmal:
@gbody: I added a couple of unit tests, then refactored the code a little bit. I added a `//$BUG` to mark what I suspect is a possible problem. If the RLE count is 0, the relocator may end up performing 2^32 relocations.

PR #835, opened 2020-02-06
VC mangled constructor should return pointer to initialised object

ptomin:
`VC` constructor (not sure if it's standard or `MSVC`-specific) returns pointer to initialised object. So modify `MsMangledNameParser` to make it work correctly


uxmal:
Let's go with this for now. Constructors are a little special, in that in C++ source, they actually don't return a value -- but their low-level implementation actually returns `this`. Thanks for the PR!

PR #836, opened 2020-02-06
Create TypedDataDumper test to reproduce truncation of 64-bit word

ptomin:
- Create `TypedDataDumper` test to reproduce truncation of 64-bit word
Expected: `dq 0x0807060504030201`
But was:  `dq 0x0000000004030201`
- Do not truncate 64-bit word (`ReadUInt32` -> `ReadUInt64`)


uxmal:
Nice catch!

PR #838, opened 2020-03-05
Fixed some null pointers when the list is empty

Piruzzolo:
maybe should be done also for SelectedItems[0]


uxmal:
Thanks for this bugfix!

uxmal:
Btw: don't forget to add yourself to `AUTHORS`!

Piruzzolo:
@uxmal I'll add myself, even if _domine, non sum dignus_

Issue #841, opened 2020-03-20
16-bit OS/2 executables are not supported


claunia:
When trying to open HPOFS20.IFS it crashes in NeImageLoader.cs;L211 because cs is 0.
When trying to open UHPOFS.DLL it crashes in LoadResources, probably because the resources format is different from Windows.

You can download the executables from https://amaus.net/static/S100/IBM/PS2/options/ in the file hpofs20.exe.

I have a NE parser in https://github.com/claunia/libexeinfo that may be of help (I also have a parser for 32-bit OS/2 executable format, LX).

uxmal:
I saw PR #842 you submitted: thanks for that. As for the LX parser, I'd be happy to accept a PR for that, as well. I will work on adding a new `OS2` platform to Reko today.

What would you say the major differences in the resource format is between Windows and OS/2 NE files?

claunia:
Hi, LX is for 32-bit excecutables, it is similar to LE, but not the worry for now.

The resource format is completely different so you need a different parser. My project libexeinfo has a parser for many of the resources. I don't remember where I got the documentation from but I tried it with several executables from OS/2 1.x.

I could not find where the entry point is defined in drivers, only found that CS points to the ring (so 0 for most drivers and 2 for some others), and that they use pascal calling convention.

I can make the resource parser while you prepare the entry point and the start of the platform and then I can add all libraries to the platform?

uxmal:
@claunia: I pushed the new ```OS2Platform``` class an hour or so ago. It's possible that OS/2 drivers have some other entry point mechanism. I'll have to hunt around for documentation for this. Adding support for the `pascal` calling convention shouldn't be too hard; it could be reused from the existing implementation in `Windows`.

claunia:
Excellent I'll take a look after lunch

claunia:
Fixed in e5636ffe701939a5b8731d7e9286ab850041a157.

uxmal:
Commit ad53150ec fixes an exception when loading the device driver file -- not sure if it's the one you reported in #842, but it may help.

PR #842, opened 2020-03-21
Partially fix loading non-Windows NE files.

claunia:
Fix loading a OS/2 NE with resources and an OS/2 driver.


uxmal:
Thanks for your first contribution! One small detail: could you add your name (and/or) alias to the AUTHORS file in the root directory?

I will address the resource types in issue #841.

src/ImageLoaders/MzExe/NeImageLoader.cs
```diff
@@ -407,17 +420,45 @@ public override Program Load(Address addrLoad)
...
+            switch (bTargetOs)
+            {
+                case NE_TARGETOS.Windows:
+                case NE_TARGETOS.Windows386:
+                    this.platform = cfgSvc.GetEnvironment("win16").Load(Services, arch);
+                    break;
+                case NE_TARGETOS.EuropeanDos:
+                    this.platform = cfgSvc.GetEnvironment("ms-dos").Load(Services, arch);
+                    break;
+                default:
```
uxmal: Commit 027904892e1a4c5e8e95b61e194d4bf241597d4d added support for the `os2` platform, you can change this to:
```C#
case NE_TARGETOS.Os2:
    this.platform = cfgSvc.GetEnvironment("os2").Load(Services, arch);
    break;
```
if you have the time.

claunia:
Ok only missing, well, whatever are you doing with icons in the Windows side. And some testing, I still get an exception somewhere.

uxmal:
Are you OK with merging, or do you wish me to wait until you've chased down the exception?

claunia:
I'm ok that way you can also take a look, you know the codebase better so you may find it first. Checked what you were doing with the Windows resources I see you are adding the Bitmap header to the DIB data in Windows resource. Not needed in OS/2 as they always have the header (but it´s a different format, I also have a decoder that gets a raw array of pixels).

uxmal:
OK I will pull it then. ¡Muchas gracias por la contribución!

claunia:
De nada. I really need to decompile that HPOFS driver so :D My benefit, your benefit, our benefit.

PR #847, opened 2020-03-22
Implement basis for LE/LX loading.

claunia:
Missing relocations, I don't know how to apply them.


uxmal:
Looks like some build files were accidentally added, under `src\Native\out`. These need to be added to `.gitignore`.

uxmal:
Looking good so far.

PR #848, opened 2020-03-24
Add more OS/2 definitions

claunia:


uxmal:
@claunia : I fixed a regression that was causing the CI builds to fail.

PR #855, opened 2020-04-28
Modify regressionTests.py to remove output files from '*.reko' dirs before running regression tests

ptomin:
Modify `regressionTests.py` to remove output files from `*.reko` directories before running regression tests

So many obsolete output files was removed


uxmal:
Thanks for cleaning that up, @ptomin .

PR #857, opened 2020-05-04
Make ExpressionTypeAscender work with nested structs

ptomin:
- Create `ExpressionTypeAscender` test with second field access of global structure variable
```
Expected:
    DataType: (ptr32 real32)
```
but was:
```
    DataType: ptr32
```
- Make `ExpressionTypeAscender.GetPossibleFieldType` work with nested structures.
If we access beyond the start of the field, we need to check if field is
nested structure

### DIFF
--- a/subjects/Hunk-m68k/BENCHLNG.reko/BENCHLNG_code.c
+++ b/subjects/Hunk-m68k/BENCHLNG.reko/BENCHLNG_code.c
@@ -23,12 +23,12 @@ void fn00001000(int32 d0, byte * a0)
 	globals->ptr3E20 = fp;
 	globals->ptr3E24 = a6_n;
 	struct Eq_n * d0_n = FindTask(0x00);
-	Eq_n a1_n = 0x12BC;
-	Eq_n d0_n = OpenLibrary(0x12BC, 0);
-	if (d0_n != 0x00)
+	struct Eq_n * a1_n = &globals->t12BC;
+	struct Eq_n * d0_n = OpenLibrary(0x12BC, 0);
+	if (d0_n != null)
 	{
 		int32 d4_n;
-		globals->t3E28 = d0_n;
+		globals->ptr3E28 = d0_n;
 		if (d0_n->ptr00AC == null)
 		{
 			d4_n = 2;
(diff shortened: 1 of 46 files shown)


subjects/Hunk-m68k/BENCHLNG.reko/BENCHLNG_code.c
```diff
@@ -23,12 +23,12 @@ void fn00001000(int32 d0, byte * a0)
 	globals->ptr3E20 = fp;
 	globals->ptr3E24 = a6_n;
 	struct Eq_n * d0_n = FindTask(0x00);
-	Eq_n a1_n = 0x12BC;
-	Eq_n d0_n = OpenLibrary(0x12BC, 0);
-	if (d0_n != 0x00)
+	struct Eq_n * a1_n = &globals->t12BC;
+	struct Eq_n * d0_n = OpenLibrary(0x12BC, 0);
```
uxmal, line 27: This a note to myself: `OpenLibrary` should have enough type info to resolve this.

subjects/PE/PPC/hello_ppc.reko/hello_ppc_text.c
```diff
@@ -1250,11 +1250,11 @@ Eq_n fn004020BC(Eq_n r2, struct Eq_n * r3, ptr32 r29, ptr32 r30, struct Eq_n * r
 				word32 ctr_n;
 				for (ctr_n = 0x05; ctr_n != 0x00; --ctr_n)
 				{
-					(&fp->aFFFFFFA0->dw0000)[r11_n] = (struct Eq_n) 0x00;
+					Mem69[fp + ~0x5F + r11_n:word32] = 0x00;
 					int32 r11_n = r11_n + 0x04;
-					(&fp->aFFFFFFA0->dw0000)[r11_n] = (struct Eq_n) 0x00;
-					(&fp->aFFFFFFA0->dw0000)[r11_n + 0x04] = (struct Eq_n) 0x00;
-					(&fp->aFFFFFFA0->dw0000)[r11_n + 0x08] = (struct Eq_n) 0x00;
+					Mem71[fp + ~0x5F + r11_n:word32] = 0x00;
```
uxmal, line 1255: Don't see a regression here. These are all minor improvements. Clearly `Mem71[fp + ~0x5F...` should be a stack variable, but that transformation should have happened during data flow analysis, so no regression.

PR #858, opened 2020-05-05
ConstantPointerAnalysis: do not add data pointer to globals if it is inside structure

ptomin:
-Create `ConstantPointerAnalysis` test with data pointer inside structure
If the data pointer is inside structure, it should be added to globals.

Expected:
```
struct (123130 struct(...) t123130)
```
but was:
```
struct (123130 struct(...) t123130) (123134 int32 dw123134) (123138 int32 dw123138)
```


uxmal:
Nice job, Pavel!

PR #859, opened 2020-05-10
Make TypedConstantRewiter work with nested structs

ptomin:
- `TypedConstantRewiterTests`: simplify `Given_Global`
Do not subtract `0x100000` from address
-  Create `TypedConstantRewriter` test with pointer to nested struct field
Expected:
```
    &globals->t100100.r0004
```
But was:
```
    &globals->dw100104
```
- Make `TypedConstantRewiter` work with complex global variables(e.g. structs)
- `ComplexExpressionBuilder.FallbackExpression`: avoid exception if data type size is unknown


subjects/PE/x86/VCExeSample/VCExeSample.reko/VCExeSample_text.c
```diff
@@ -125,10 +125,10 @@ void gbl_nested_structs_test14()
...
 	globals->gbl_nested_structs = (Eq_n) 0x05;
 	// gbl_nested_structs.str.b = 6
-	globals->t403020 = (Eq_n) 0x06;
+	globals->gbl_nested_structs.str.b = (Eq_n) 0x06;
 	// gbl_nested_structs.str.c = 7
-	globals->dw403024 = 0x07;
+	globals->gbl_nested_structs.str.c = 0x07;
 	// gbl_nested_structs.d = 8
-	globals->dw403028 = 0x08;
+	globals->gbl_nested_structs.d = 0x08;
```
ptomin, line 132: Look better, is not it?

src/Decompiler/Typing/TypedConstantRewriter.cs
```diff
@@ -365,6 +363,39 @@ DataType PromoteToCString(Constant c, DataType charType)
             return dt;
         }
 
+        private bool IsInsideField(int offset, StructureField field)
+        {
+            if (offset == field.Offset)
+                return true;
+            if (offset < field.Offset + field.DataType.Size)
+                return true;
+            var str = field.DataType.ResolveAs<StructureType>();
```
uxmal, line 372: Should this not be:
```
if (field.Offset <= offset && offset < field.Offset + field.DatType.Size)
```

ptomin: `StructureFieldCollection.LowerBound`  guarantees that `field.Offset <= offset` is always `true`. But I'll add additional check for safety

src/Decompiler/Typing/TypedConstantRewriter.cs
```diff
@@ -365,6 +363,41 @@ DataType PromoteToCString(Constant c, DataType charType)
             return dt;
         }
 
+        private bool IsInsideField(int offset, StructureField field)
+        {
+            if (offset < field.Offset)
+                return false;
```
ptomin, line 369: John, is this check enough?

uxmal: Let's go with this.

PR #861, opened 2020-05-17
ComplexExpressionBuilder: correct type for '&' expression

ptomin:
When rewriting `expression + offset` to `&complexExpression`, data type
of result should be pointer to `complexExpression.DataType`,
not `expression.DataType`

### DIFF
--- a/subjects/regressions/reko-90/PP.reko/PP_0800.c
+++ b/subjects/regressions/reko-90/PP.reko/PP_0800.c
@@ -12153,7 +12153,7 @@ l0800_nFCB:
 		si_n = ϕ(si_n, si_n, si_n);
 		Mem53 = ϕ(Mem968, Mem847, Mem49);
 		al_n = (ds->*si_n).b0000;
-		si_n = &&(SLICE(ptrArg02, selector, 16)->*(SLICE(ptrArg02, selector, 16)->*si_n).b0001);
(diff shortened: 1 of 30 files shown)


subjects/regressions/reko-90/PP.reko/PP_0800.c
```diff
@@ -12153,7 +12153,7 @@ Eq_n fn0800-8F97(Eq_n di, Eq_n ds, <anonymous> Eq_n::* wArg02, <anonymous> Eq_n:
 		si_n = ϕ(si_n, si_n, si_n);
 		Mem53 = ϕ(Mem968, Mem847, Mem49);
 		al_n = (ds->*si_n).b0000;
-		si_n = &&(SLICE(ptrArg02, selector, 16)->*(SLICE(ptrArg02, selector, 16)->*si_n).b0001);
+		si_n = (struct Eq_n Eq_n::*) &&(SLICE(ptrArg02, selector, 16)->*(SLICE(ptrArg02, selector, 16)->*si_n).b0001);
```
uxmal, line 12156: The '&&' is problematic here, but existed before the changes in the PR.  Because this particular function is failing decompilation -- see the  ϕ's that weren't stripped -- all bets are off. We should try to fix the decompilation of `fn0800_8F97`, but not as part of this PR.

uxmal:
Thanks for the PR, Pavel.

PR #862, opened 2020-05-17
TypedConstantRewriter: correct type for '&' expression

ptomin:
The same fix as #861


subjects/Elf/Msp430/a.reko/a_text.c
```diff
@@ -399,12 +399,12 @@ ui20 vTaskDelayUntil(struct Eq_n * pc, ui20 sr, Eq_n r14, Eq_n r15)
 		globals->t020A = r11_n;
 		if (r11_n >= 0x0208)
 		{
-			r14_n = &globals->t020A;
+			r14_n = (struct Eq_n *) &globals->t020A;
```
uxmal, line 402: Is this a regression?

subjects/Hunk-m68k/BENCHFN.reko/BENCHFN.h
```diff
@@ -24861,7 +24861,7 @@ T_5544: (in 0<16> : word16)
   Class: Eq_5537
   DataType: word16
   OrigDataType: word16
-T_5545: (in __btst((&(&(&(&(&(&(&(&(&(&(&(&(&(&(&(&(&(&(&(&globals->t288D)[(int32) (int16) (d0_2124 & 0xFF<32>)].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>].a0000)[0<i32>], 0<16>) : bool)
+T_5545: (in __btst((&globals->t288D)[(int32) (int16) (d0_2124 & 0xFF<32>)], 0<16>) : bool)
```
uxmal, line 24864: The deeply nested `&(&(&...` seems to go away here. This is an improvement!

subjects/Elf/Msp430/a.reko/a_text.c
ptomin: I'll do investigation this week.

ptomin: `0x202<p16>` has incompatible type with pointer to `globals->t020A` after previous stages of type analysis
﻿And maybe `mov.w &0202,r15` incorrectly rewritten to `r15 = 0x202<p16>`. Shouldn't be `r15 = Mem[0x202<p16>]`?

ptomin:
> Do you know the reason why?

I'm not sure but maybe incorrect rewriting could cause it

uxmal:
Ok i will review the manual to make sure it isn't a rewriter error.

subjects/Elf/Msp430/a.reko/a_text.c
uxmal: You're right; I've checked with the MSP430 manual. The `&0202` stands for "absolute address", and should indeed be translated to `Mem[0x0202<p16>:...]`. Working on a fix now. PR can probably go ahead without waiting for my fix however.

uxmal:
I'm going ahead with the PR. Nice work uncovering the rewriter error, Pavel. I will fix that separately.

PR #866, opened 2020-05-26
Minor cleanups

blindmatrix:
- Removed code that seams to be dead (program.User.RegisterValues should never be null)
- Removed some compile time warnings from Native\ArmNative


uxmal:
Nice! Let's whittle down those warnings...

PR #869, opened 2020-05-28
Don't append <nBits> when saving registerValues

blindmatrix:
- Removed <nBits> from the saved project for registerValues
- Delay opening the output file until the serialization is complete (prevents empty files if anything throws)


uxmal:
Looks good. Nice work!

PR #872, opened 2020-06-06
Eliminate fictitious "global" identifiers.

ptomin:
For instance, transform `global->ptr0001` to `g_ptr0001` or
`global->userVariable` to `userVariable`


ptomin:
I have found 3 regression tests that never run during testing this feature:
https://github.com/uxmal/reko/tree/4603b2f6b49a6b4eb312530b645a05d1e47b716e/subjects/Elf/MIPS/animals
https://github.com/uxmal/reko/tree/4603b2f6b49a6b4eb312530b645a05d1e47b716e/subjects/PE/x86-64/load
https://github.com/uxmal/reko/tree/4603b2f6b49a6b4eb312530b645a05d1e47b716e/subjects/regressions/reko-15

src/Decompiler/Structure/FictitiousGlobalEliminator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public void Transform()
+        {
+            foreach (var stm in proc.Statements)
+            {
+                stm.Instruction = Transform(stm.Instruction);
+            }
+        }
+
+        public override Expression VisitFieldAccess(FieldAccess acc)
```
uxmal, line 52: The logic is fine, but do you agree that it makes sense to add this logic to `ComplexExpressionBuilder` and `TypedConstantRewriter` instead? That would avoid injecting references to the temporary `global` in the first place.

ptomin: Done in 8ae7402

uxmal:
Nice initiative, Pavel. I may visit these classes and make some small changes but the logic is sound. The code looks much better -- and more correct.

PR #876, opened 2020-06-12
Do not throw exception when infinite recursion detected during Type Analysis

ptomin:
This is the work around infinite recursion which makes possible
to continue `Type Analysis` after such errors

### DIFF
--- a/src/Core/Types/Unifier.cs
+++ b/src/Core/Types/Unifier.cs
@@ -78,8 +78,11 @@ namespace Reko.Core.Types
 			if (a == null || b == null)
 				return false;
 
-			if (depth > 20)
-				throw new StackOverflowException("Way too deep");     //$BUG: discover why datatypes recurse so deep.
+            if (depth > 20)
+            {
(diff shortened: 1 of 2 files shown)


src/Core/Types/Unifier.cs
```diff
@@ -78,8 +78,11 @@ private bool DoAreCompatible(DataType a, DataType b, int depth)
 			if (a == null || b == null)
 				return false;
 
-			if (depth > 20)
-				throw new StackOverflowException("Way too deep");     //$BUG: discover why datatypes recurse so deep.
+            if (depth > 20)
+            {
+                DebugEx.Error(trace, "Way too deep");     //$BUG: discover why datatypes recurse so deep.
```
uxmal, line 83: prefer `trace.Error(...)`. `DebugEx.Error` is an extension method.

ptomin: Done. See https://github.com/uxmal/reko/compare/f885e6d9ccd7c6e32939b45976fc8f84276a916b..180d2a69f84d4a67c41f0ae3cab5a7c5d393ffc3

PR #877, opened 2020-06-15
Feature: Simplify x86 "not" sequence

blindmatrix:
Not sure if this is the correct place for this fix, but at least it works.

`neg ax`
`sbb ax, ax`
`inc ax`

results in code like

`0x00 - (-MyVariable == 0x00) + 0x01`

This will convert it to

`!MyVariable`

The sequence might be x86 specific, but the pattern it searchers for could be valid for other architectures too. But it might need a generalization to match other sources. This was brute-force ;)

ps. Should have named the branch "ComplicatedNotSimplified" instead...


src/Decompiler/Evaluation/ComplicatedNotRule.cs
```diff
@@ -0,0 +1,86 @@
...
+    ///
+    /// Example x86 assembler:
+    /// neg ax
+    /// sbb ax, ax
+    /// inc ax
+    ///
+    /// Example result:
+    /// turn func(0x00 - (-MyVariable == 0x00) + 0x01) into func(!MyVariable)
+    /// </summary>
+    public class ComplicatedNotRule
```
uxmal, line 41: The naming is an issue; it's conceivable even more complicated `Not` rules might be needed. Naming is hard. How about `LogicalNotFromAdditionRule`? The initial expression is forming a logical not from the additions.

src/Decompiler/Evaluation/ComplicatedNotRule.cs
```diff
@@ -0,0 +1,86 @@
...
+            if (!leftExpression.Left.IsZero)
+                return false;
+
+            if (!(leftExpression.Right is BinaryExpression middleExpression) || middleExpression.Operator != Operator.Eq)
+                return false;
+
+            if (!middleExpression.Right.IsZero)
+                return false;
+
+            if (!(middleExpression.Left is UnaryExpression unaryExpression) || unaryExpression.Operator != Operator.Neg)
```
uxmal, line 72: Consider supporting `0 - (exp == 0) + 1` as well as `0 - (-exp == 0) + 1`. That is, change this logic to:
```C#
if (middleExpression.Left is UnaryExpression unary && unaryExpression.Operator == Operator.Neg) 
{
    expression = unaryExpresion.Expression;
}
else 
{ 
    expression = middleExpression.Left;
}
```

blindmatrix: I'll look into it later tonight, good point! The reason I only supported -exp is that this is how they setup `sbb` to _do it's thing_ (the part about x86 specific). Do you have any samples that is affected by this that I can look at too, to make sure it's correct? Otherwise I'll just look into your example and make sure my case still works :P

src/Decompiler/Evaluation/ComplicatedNotRule.cs
blindmatrix: Sure thing! I'll look into it later.

blindmatrix: Unless you say something else I'm going with `LogicalNotFromBorrowRule` as that's what I expect the majority of sources will be. Other platforms can probably create something else, but unless it's done with the borrow-magic I would guess they produce some other formula.

You can give suggestions here, or on next push :P But I think that this name is more informative.

src/Decompiler/Evaluation/ExpressionSimplifier.cs
```diff
@@ -389,6 +391,18 @@ public virtual Expression VisitBinaryExpression(BinaryExpression binExp)
...
+                Changed = true;
+                return complicatedNot.Transform();
+            }
+
+            if (complicatedNot.Match(binExp))
+            {
+                Changed = true;
+                return complicatedNot.Transform();
+            }
+
```
ptomin, line 405: Duplicated call of `complicatedNot` rule

blindmatrix: Thanks, I noticed that too, it's fixed on the upcoming push :)

src/Decompiler/Evaluation/ComplicatedNotRule.cs
uxmal: I'm seeing _similar_ code sequences in ~/subjects/Elf/x86-65/ls: for instance, the basic block `l000000000040A545` looks like this:
```
	and	edi,01
	mov	byte ptr [rsi+0A],20
	mov	byte ptr [rsi+0B],00
	cmp	edi,01
	sbb	eax,eax
	and	eax,E0
	add	eax,74
	mov	[rsi+09],al
	ret
```
Clearly a variant of the "subtract carry flag from register to yield either all 0's or all 1's" trick is being used here to generate either a lower case 't' (U+0074) or an upper case 'T' (U+0054). But this is not the same as your example.

In ~/subjects/MZ-x86/BENCHFN.EXE, I see in basic block `l0800_0A87` the following code:
```
	neg	cx
	sbb	ax,ax
	push	ax
```
which is _similar_ but again not the same as your sample.

~/subjects/PE/x86/ExeDll/Executable.exe has the end of basic block `l0040178F` looking like this:
```
; preceding instructions deleted
	neg	ebx
	lea	eax,[ebp-58]
	mov	[ebp-08],eax
	lea	eax,[ebp-00000324]
	sbb	bl,bl
	mov	[ebp-04],eax
	inc	bl
	call	dword ptr [00402038]
```
Notice the interleaving of instructions before the `call` instruction. Reko is currently (without your fix) generating this:
```
	byte bl_90 = 0<8> - (IsDebuggerPresent() == 1<32>)
```

src/Decompiler/Evaluation/ComplicatedNotRule.cs
uxmal: I'm fine with that name.

src/Decompiler/Evaluation/LogicalNotFollowedByNegRule.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using Reko.Core;
+using Reko.Core.Expressions;
+using Reko.Core.Operators;
+using Reko.Core.Types;
+using System;
+
+namespace Reko.Evaluation
+{
+    class LogicalNotFollowedByNegRule
```
uxmal, line 29: Please add a clarifying comment here to explain what the class does.

src/Decompiler/Evaluation/UnaryNegEqZeroRule.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using Reko.Core;
+using Reko.Core.Expressions;
+using Reko.Core.Operators;
+using Reko.Core.Types;
+using System;
+
+namespace Reko.Evaluation
+{
+    public class UnaryNegEqZeroRule
```
uxmal, line 29: Please add a clarifying comment here. We should also be quite explicit about the code assuming that two's complement representation is being used here. Who knows what -0 means on one's complement machines.

blindmatrix:
I have to run the regressions too. I only ran the UnitTests so far.

src/Decompiler/Evaluation/ComplicatedNotRule.cs
blindmatrix: I removed that completely from this part and added two new rules, one simplifying unary expressions of the type (!-exp) into (!exp) and a binary that convert (-exp == 0) into (exp == 0). I don't believe I've seen a platform where the sign of an integer changes if it evaluates to zero or not.

The only thing that I could think of if the hardware in question would raise an exception if a value overflowed when changing the sign (like abs(MAX_INT) < abs(MIN_INT)). But I think other simplifications would ruin that too, so I don't think it's a loss.

uxmal:
Looks good. Thanks for the contribution!

blindmatrix:
Thanks m8 ;)

ptomin:
@blindmatrix Nice work to simplification
```
           neg eax
           sbb eax, eax
           neg eax
```
sequence to `not eax`!

Issue #881, opened 2020-06-21
x86-16 Incorrect parameters used when using return-values as parameters (faked far calls)


blindmatrix:
Wrong stack location is used when making function calls where some arguments are the result of other function calls when the "argument function" is a far proc that's called as a near call (`push cs; call near func;`)

Affected assembly:
```asm
; func and argfunc1 are far procs
push 4                ; sp+A when inside func
push 2                ; sp+8 when inside func
push cs               ; This is the problem (extra sp = sp-2)
call near argfunc1
push ax               ; sp+6 when inside func
push 1                ; sp+4 when inside func
push cs
call near func
```

Reko fails to remove the added offset for the first `push cs` when valuating the arguments for `func`.

Demo program: [pushcs.zip](https://github.com/uxmal/reko/files/4810263/pushcs.zip)

```C
volatile int var1;
volatile int var2;

void func(int a, int b, int c, int d)
{
	var2 = a + b + c + d;
}

int argfunc1()
{
	return var1;
}

int main(void) {
	func(1, argfunc1(), 3, 4);
	
	return 0;
}
```

Expected output:
```C
void main()
{
    func(0x01, argfunc1(), 0x03, 0x04);
}
```

Actual output:
```C
void main()
{
    func(0x01, argfunc1(), wLoc06, 0x03);
}
```

I believe the issue is in how the rewriter handles `push cs; call near func;` in `X86Rewriter.RewritePush`

blindmatrix:
Having a quick look I think even I can fix this :P I'm going to give it a try.

uxmal:
See my comments on #882. I have a fix on my machine, but there are some regressions in the regression suite I need to examine before I can push to `master`.

uxmal:
My fix wasn't working correctly, and the likelihood of `push cs / call <near>` not being a simulated far call is so small that I decided to pull #882 instead. Thanks for taking the time to report this.

PR #882, opened 2020-06-21
x86-16 fix rewrite of push cs; call near func; by converting it to a real far call

blindmatrix:
I didn't add any new tests but updated old once. There where 2 unit tests and a regression that failed. The regression I've read and I can't see anything that looks incorrect, but a second pair of eyes on the regression outputs would be appreciated.

This will only fix calls that abuse near relative calls prefixed with `push cs`. I have yet to see this being used with a `push cs` followed by an indirect near call. It might exist, but it's nothing I've seen so far.

I ran the regressions with Python3.

Also, there was a failure in `subjects/PE/x86/RussianText/RussianText.reko/RussianText.globals.c` but that also happens before this change, so I'm ignoring that :P

So, tests will fail until AppVeyor run Python3 or regressionTests.py gets fixed for Python2, and RussionText is solved.

### DIFF
--- a/src/Arch/X86/X86Rewriter.Alu.cs
+++ b/src/Arch/X86/X86Rewriter.Alu.cs
@@ -744,8 +744,18 @@ namespace Reko.Arch.X86
                     dasm.Peek(1).Operands[0].Width == PrimitiveType.Word16)
                 {
                     dasm.MoveNext();
-                    m.Assign(StackPointer(), m.ISubS(StackPointer(), reg.Register.DataType.Size));
-                    RewriteCall(dasm.Current.Operands[0], dasm.Current.Operands[0].Width);
+                    MachineOperand targetOperand = dasm.Current.Operands[0];
+
+                    if (targetOperand is ImmediateOperand immOperand)
+                    {
+                        targetOperand = new X86AddressOperand(orw.ImmediateAsAddress(instrCur.Address, immOperand));
+                    }
+                    else
+                    {
+                        m.Assign(StackPointer(), m.ISubS(StackPointer(), reg.Register.DataType.Size));
+                    }
+
+                    RewriteCall(targetOperand, targetOperand.Width);
                     this.len = (byte)(this.len + dasm.Current.Length);
                     return;
                 }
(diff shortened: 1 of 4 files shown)


blindmatrix:
Maybe I should add a specific test after all too... I'll look into it at lunch tomorrow

uxmal:
I'm not convinced that always treating a pushed `cs` register as part of the return address is a safe assumption, although clearly it is used frequently to make near calls to far procedures in the same segment as the caller. I'm concerned this might introduce problems for code which passes a segment selector as a parameter on the stack.

Reko should be recording, for each `Procedure`, how it affects the stack as it returns to its caller. In the `RtlReturn` class there is a `ReturnAddressBytes` property, which should be getting propagated to the `StackDelta` property of the `Procedure.Signature` instance during the scanning phase. There are known problems with this, but it should be working for most procedures. I will look at #881 and see if I can make some changes that improve the generated output.

blindmatrix:
I have opinions :P

I have some suggested paths forward.

The best approach (I think) would be to mark procedures as near/far/undefined when scanning (or based on CSignature). If the procedure is marked as far, but called using `push cs; call near;`, then this should be applied. There should of course be an option to override this, globally and/or per call site.

The second approach (which probably is a bit lazy) would be to assume this is always correct, but again with an option to override globally and or locally.

I can see the point of tracking StackDelta too, but if you know the target is _far_ but called with this setup, then I think it's clearer to let the Rewriter produce the intended semantics.

However, if you had instructions between the _push_ and the _call_ then my approaches would fail. This could be handled later as a pass, but I have yet to see that in a real binary, as this is a specific pattern produced by some compilers.

But I'm curious about a sample where `push cs; call near;` is actually used but not targeting a _far_ procedure.

Also as a side note, I don't think the current patch would break more then the existing code. As far as I can see only SP gets updated, CS never get stored in the current output, meaning no existing decompilation should (hopefully) break more then I think it might already be :P

uxmal:
I went ahead and pulled the PR. It's unlikely to see the sequence `push cs / call <near>` and it isn't a call to a `far` procedure. Thanks for the PR and patience!

PR #886, opened 2020-06-24
Fix regressionTests.py to work with Python2 again

blindmatrix:
regressionTests.py work with Python2 again.

subjects/PE/x86/RussianText/RussianText.reko/RussianText.globals.c is incorrect so this will still break with Python3.
It declares a _char_ that's really the first half of a 2-byte utf-8 sequence and Python3 care about characters vs bytes.


uxmal:
Thanks for this -- it had gone me by since I only have Python 3 on my dev box. May be a good time to ensure AppVeyor also uses Python 3.

PR #888, opened 2020-06-27
Read external procedure characteristics for user-defined metafiles

ptomin:
- Move Python characteristics to user defined metadata (`python.xml`)
- Read external procedure characteristics from user-defined metafiles. Modify `TypeLibrary` to store procedure characteristics


uxmal:
Looks good Pavel, in it goes.

PR #889, opened 2020-06-28
Read characteristics for mangled procedures

ptomin:
- Create `DynamicLinker` test for mangled procedure characteristics
Expected:
```
void foo()
VarargsParserClass: FakeParser
```

but was:
```
void foo()
VarargsParserClass:
```
- Read characteristics for mangled procedures. For instance, we have `?text_format@@YAPBDPBDZZ`procedure. It will be
`char const * text_format(char const *, ...)` after demangling. It would be nice to set `vararg` parser for it. So modify `DynamicLinker.ResolveProcedure` to read procedure characteristics


uxmal:
Looks good Pavel.

PR #890, opened 2020-06-28
Fixed incorrect caching in TypeTransformer, and added Unions to the cache.

blindmatrix:
A HashSet was used to mark visited StructureType instances. (I'm guessing, but I believe) for two reasons, first would be performance as an already processed instance should be left as-is, but also to prevent a loop if a structure would contain a pointer to itself. Not sure if one of the above or both.

I changed that to a Dictionary as (if it was used as a cache) it missed the case when the entire structure was replaced (and only handled the case where its fields was transformed).

I also added Union to this set.

This would make it a lot less likely to hit the "50 iterations" limit. It also gave me a massive decrease in time for a particular project (10+ minutes ending in hitting the 50 iterations limit), and modest increases for smaller projects.

This patch change a lot of text in the regressions but all the structures/unions look to have the same shape and size (as to the best of my abilities) and a few _struct_-entries got converted into _typedef_-entries instead. 

I believe the actual problem that was solved was when you had a layout like this

```C
struct ThisIsTheProblem {
        word16 singleField;
};

struct NotAVictime
{
        ThisIsTheProblem aField;
};

struct ThisIsAVictime1
{
        ThisIsTheProblem aField;
};

struct ThisIsAVictime2
{
        ThisIsTheProblem aField;
};
```

`ThisIsTheProblem` would be simplified into `word16`, but that would only happen for `NotAVictime`. `ThisIsAVictime1` and `ThisIsAVictime2` would still have `ThisIsTheProblem` as the HashSet would have marked it has "handled" already, but would have failed to give the real outcome that `NotAVictime` would have gotten.

### DIFF
--- a/src/Decompiler/Typing/TypeTransformer.cs
+++ b/src/Decompiler/Typing/TypeTransformer.cs
@@ -43,7 +43,7 @@ namespace Reko.Typing
         private DecompilerEventListener eventListener;
 
 		private static TraceSwitch trace = new TraceSwitch("TypeTransformer", "Traces the transformation of types") { Level = TraceLevel.Verbose };
-        private HashSet<DataType> visitedTypes;
+        private readonly Dictionary<DataType, DataType> visitedTypes;
 
         public TypeTransformer(TypeFactory factory, TypeStore store, Program program)
             : this(factory, store, program, new NullDecompilerEventListener())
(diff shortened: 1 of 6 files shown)


uxmal:
Thanks for looking into this. There are a few places where cycles in the type graph are still derailing the `TypeTransformer` but they're due to other parts of the code.

PR #891, opened 2020-06-28
regressionTests.py sometime failed when output files were locked

blindmatrix:
So I though I fixed the race condition in regressionTests.py, but I got it again today. The things I said about renaming before deletes in Windows is still true, but the chance of it happening is very small... I instead think that the output files might be locked by either Windows Search Indexer or a malware scanner.

I reverted the rename/remove thing I did last time and instead added a retry after a delay. The race condition is still there, but this time it should be less likely to cause an issue.

### DIFF
--- a/subjects/regressionTests.py
+++ b/subjects/regressionTests.py
@@ -82,17 +82,23 @@ def cmdline_split(s):
         a.append(sub)
     return a
 
-# On Windows file removal is asynchronous but renames are atomic
-def remove_file(file):
-    os.rename(file, '%s.rmtmp' % file)
-    os.remove('%s.rmtmp' % file)
-
 # Remove output files
 def clear_dir(dir_name, files):
+    failedFiles = []
     for pname in files:
         for ext in output_extensions:
             if pname.endswith(ext):
-                remove_file(os.path.join(dir_name, pname))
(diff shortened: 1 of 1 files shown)


subjects/regressionTests.py
```diff
@@ -82,17 +82,23 @@ def cmdline_split(s):
...
+    failedFiles = []
     for pname in files:
         for ext in output_extensions:
             if pname.endswith(ext):
-                remove_file(os.path.join(dir_name, pname))
+                filename = os.path.join(dir_name, pname)
+                try:
+                    os.remove(filename)
+                except:
+                    failedFiles.append(filename)
```
uxmal, line 95: Let's add a comment here like
`# Malware scanners may block the deletion, so try again later.`

uxmal:
There are many reasons why the files may be blocked, so this seems like a reasonable fall back. I'm adding a request for a comment explaining why, then this is good to go.

blindmatrix:
I added a comment about this, do you want it? https://github.com/blindmatrix/reko/commit/1692ea15c4b67f2666865bf52a075f2e3ad6bfde

uxmal:
It's OK, I added it myself. Thanks for chasing this so persistently!

PR #909, opened 2020-07-06
Fix naming of 'scanf' format parser at windows characteristics files

ptomin:
Fix naming of `scanf` format parser at windows characteristics files. It caused exception during scanning phase.


uxmal:
I could have sworn I fixed this. Thanks for fixing.

PR #910, opened 2020-07-07
Added "Open in new tab" to context menu in CombinedCodeViewInteractor

blindmatrix:
Added a menu item for the context menu in CombinedCodeViewInteractor that opens the selected function in a new tab.

Not sure if this is the correct way to do it but "it worked on my machine<sup>TM</sup>"


blindmatrix:
I'm not sure I feel my change caused the code complexity failure, what do you think?

uxmal:
I think CodeFactor is being a little overzealous here. The code look fine.

PR #911, opened 2020-07-09
Create ExpressionSimplifier test: '(real32)0x1'

ptomin:
John, I do not agree with simplification at 4971642. `Reko.Core.Expressions.Cast` is C-style cast and can't be used as reinterpret cast. So we should not transform `(real32) 1<32>` to `1.4013e-45`. It should be `1.0` instead
I have create `ExpressionSimplifier` test: '(real32)0x1
Expected:
```
    1.0F
```
but was:
```
    0.0F
```

### DIFF
--- a/src/UnitTests/Evaluation/ExpressionSimplifierTests.cs
+++ b/src/UnitTests/Evaluation/ExpressionSimplifierTests.cs
@@ -368,6 +368,19 @@ namespace Reko.UnitTests.Evaluation
         }
 
         [Test]
+        public void Exs_Word32ConstantToReal32Convert()
+        {
+            Given_ExpressionSimplifier();
+            var value = Constant.Word32(0x1);
+            var exp = m.Convert(
+                value, PrimitiveType.Word32, PrimitiveType.Real32);
+
+            var result = exp.Accept(simplifier);
+
+            Assert.AreEqual("1.0F", result.ToString());
+        }
+
+        [Test]
         public void Exs_SegMem_Constants()
         {
             Given_SegmentedArchitecture();


ptomin:
It should be postponed to finish implementation of #913

ptomin:
I have modified test to use `Conversion` instead of `Cast`

uxmal:
This highlights a problem: the x86 rewriter should not be emitting conversions from `word32` to `real32`, as they make no sense. It should be emitting conversions from `int32` to `real32` / `real64` as appropriate. I will go through the various rewriters to make sure that when integer to/from floating point conversions are happening, the non-floatingpoint type parameter is always either `intxx` or `uintxx`, but not `wordxx`.

Note that sign and zero extensions also are modeled with `convert`. However the extensions are OK do to on `wordxx` types as sign or zero extensions imply no interpretation of the bit pattern in a word. A conversion to/from `real` is in a sense "stronger" as it implies actual data types on the operands (2's complement / IEEE floating point in the current state of Reko).

ptomin:
Agree. Word could be real, pointer, integer. So `Convert(e, word<xx>, real<xx>` is incorrect. Maybe `Conversion` constructor should be modified to throw exception if word-to-real conversion is emitted.
And https://github.com/uxmal/reko/blob/a822a0a45e523cb099ed1d37e48964ea96859260/src/Decompiler/Evaluation/ExpressionSimplifier.cs#L522-L526 should be removed

uxmal:
This has been fixed on my machine. I'm doing some other refactorings and fixes as well, but will merge with this PR soon.

PR #912, opened 2020-07-12
Correct FPU return binding for user-defined procedures

ptomin:
- Add function which returns `double` to `VCExeSample`
Expected:
```
    return d;
```
but was:
```
    return <invalid>;
```
- Correct `FPU` return binding: `-1` offset instead of `0`


uxmal:
My apologies for taking so long with this review, @ptomin. This is an improvement on the existing codebase.

Issue #914, opened 2020-07-12
System.OverflowException: Arithmetic operation resulted in an overflow.


ptomin:
Reko throws exception during analysis phase on my binaries:
```
error: An error occurred during decompilation. System.OverflowException: Arithmetic operation resulted in an overflow.
at System.Convert.ToUInt64(Double value)
at Reko.Core.Expressions.ConstantReal64.ToUInt64() in d:\dev\uxmal\reko\master\src\Core\Expressions\Constant.cs:line 1757
at Reko.Evaluation.SliceConstant_Rule.Transform() in d:\dev\uxmal\reko\master\src\Decompiler\Evaluation\SliceConstant_Rule.cs:line 45
at Reko.Evaluation.ExpressionSimplifier.VisitSlice(Slice slice) in d:\dev\uxmal\reko\master\src\Decompiler\Evaluation\ExpressionSimplifier.cs:line 871
at Reko.Core.Expressions.Slice.AcceptT in d:\dev\uxmal\reko\master\src\Core\Expressions\Slice.cs:line 64
at Reko.Analysis.ValuePropagator.VisitAssignment(Assignment a) in d:\dev\uxmal\reko\master\src\Decompiler\Analysis\ValuePropagator.cs:line 103
at Reko.Core.Code.Assignment.AcceptT in d:\dev\uxmal\reko\master\src\Core\Code\Assignment.cs:line 49
at Reko.Analysis.ValuePropagator.Transform(Statement stm) in d:\dev\uxmal\reko\master\src\Decompiler\Analysis\ValuePropagator.cs:line 95
at Reko.Analysis.ValuePropagator.Transform() in d:\dev\uxmal\reko\master\src\Decompiler\Analysis\ValuePropagator.cs:line 86
at Reko.Analysis.DataFlowAnalysis.ConvertToSsa(Procedure proc) in d:\dev\uxmal\reko\master\src\Decompiler\Analysis\DataFlowAnalysis.cs:line 490
```
I suggest to introduce new method `int64 Constant.ToBitPattern()` which does reinterpret cast (e. g. for `1.0` returns `0x3ff0000000000000` instead of `0x1`) and use it instead of `ToInt64` in `ExpressionSimplifier`

uxmal:
Could you hunt down the expression that exhibits this behavior? Let's put it under unit test and see how to fix it now that `Convert` expressions are available.

ptomin:
OK. I'll try to create unit test

ptomin:
[...]

PR #936, opened 2020-08-31
Create ExpressionSimplifier unit test for slice of large double value

ptomin:
Create `ExpressionSimplifier` unit test for slice of large double value
`System.OverflowException` exception was thrown


ptomin:
This unit test can help to reproduce bug described in #914

PR #948, opened 2020-10-03
Scanner: enqueue procedures obtained from DataScanner [Bug fix for Disabled Shingle scanner for pySample]

ptomin:
Scanner loses procedure body if `Shingle` scanner is disabled. It looks like `GlobalDataWorkItem` was broken during new `Scanner` implementation.


ptomin:
I have modified `Scanner` to enqueue procedures obtained from DataScanner

uxmal:
What happens if you turn on the shingle scanner again?

ptomin:
I have created `shingledPySample.dcproject`. See output: https://github.com/ptomin/reko/tree/fec8de2d2a7515bd4f69d85b0ca9157537b72ab1/subjects/PE/x86/pySample/shingledPySample.reko
Now we have two projects. First one has disabled Shingled scanner, and it is enabled for second one

uxmal:
I'm look at this a little deeper.

uxmal:
Pavel, I looked at `shingledPySample.reko\pySample_text.c`, and I cannot find a procedure with missing body. Do you remember which procedure was having problems?

subjects/PE/x86/pySample/pySample_text.c
```diff
@@ -7,37 +7,21 @@
...
-	return eax_n;
 }
 
 // 10001050: Register (ptr32 Eq_n) fn10001050(Stack (ptr32 Eq_n) ptrArg04, Stack (ptr32 Eq_n) ptrArg08)
 PyObject * fn10001050(PyObject * ptrArg04, PyObject * ptrArg08)
 {
-	PyObject * eax_n = PyArg_ParseTuple(ptrArg08, "ii:dif", fp - 0x08, fp - 0x04);
-	if (eax_n != null)
-		return Py_BuildValue("i", dwLoc08 - dwLoc04);
-	return eax_n;
```
ptomin: There is no procedure body

ptomin:
>Pavel, I looked at shingledPySample.reko\pySample_text.c, and I cannot find a procedure with missing body. Do you remember which procedure was having problems?

See https://github.com/uxmal/reko/pull/948/commits/fe3c18fc2e98a19ecf24fd88bf2575fd27d8c873#r500292692. There was problem after fe3c18f. I have fix it in 05e12b7. So it shouldn't be reproduced now

PR #976, opened 2020-11-14
x86: rewriters for por/vpor

smx-smx:


uxmal:
I'm accepting this -- the recent upgrade to VS2019 seems to have hosed the CI builds.

PR #985, opened 2020-12-09
Small readme change to avoid confusion.

claunia:
Some people would think it only decompiles C#, instead of thinking it is written in C#.


uxmal:
Thanks for this. I agree that mentioning the implementation language so soon might be confusing.

PR #991, opened 2020-12-27
Make possible to use reprocessed C headers as metafiles for Reko project

ptomin:
Create loader for preprocessed C header


uxmal:
Great work! I had been mulling over this for a long while, thanks for making it happen.

PR #997, opened 2021-02-13
Create sample executable with stack allocated variables

ptomin:
Add sample which contains memory allocation at stack frame for data structures.
- `Reko` loses definitions of field values for such structures.
- There is too aggressive propagation of field values (e.g. should be `printf("%d %f %d %f\n", s1.i, s1.d, s2.i, s2.d)`  instead of `printf("%d %f %d %f\n", 100, 1.0, 0x0A, rLoc1_n)`)


uxmal:
Makes sense to have a small binary for this purpose. Already I see that `Eq_8` and Eq_11` need to be unified.

PR #999, opened 2021-02-21
netStandard: generate .msi installer on Windows

smx-smx:


uxmal:
Thanks for all the hard work in getting this up and running.

PR #1001, opened 2021-02-28
Win16Platform: Support __ prefix for calling conventions

blindmatrix:
Allow explicit __pascal calling convention in signatures


uxmal:
Thanks for the contribution!

PR #1003, opened 2021-03-03
Mixed code data model null pointer exception

blindmatrix:
MixedCodeDataModel threw a NullPointerException when hovering over references because some fields wasn't copied in the copy constructor.

Also marked read only fields as readonly.

### DIFF
--- a/src/UserInterfaces/WindowsForms/Controls/MixedCodeDataModel.cs
+++ b/src/UserInterfaces/WindowsForms/Controls/MixedCodeDataModel.cs
@@ -37,12 +37,12 @@ namespace Reko.UserInterfaces.WindowsForms.Controls
     {
         const int BytesPerLine = 16;
 
-        private Program program;
-        private ImageMap imageMap;
+        private readonly Program program;
+        private readonly ImageMap imageMap;
         private ModelPosition curPos;
-        private ModelPosition endPos;
-        private Dictionary<ImageMapBlock, MachineInstruction[]> instructions;
-        private IDictionary<Address, string[]> comments;
+        private readonly ModelPosition endPos;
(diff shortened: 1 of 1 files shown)


src/UserInterfaces/WindowsForms/Controls/MixedCodeDataModel.cs
```diff
@@ -164,9 +167,9 @@ private static Address Align(Address addr, uint alignment)
         /// Preemptively collect the machine code instructions
         /// in all image map blocks.
         /// </summary>
-        private void CollectInstructions()
+        private Dictionary<ImageMapBlock, MachineInstruction[]> CollectInstructions()
         {
-            this.instructions = new Dictionary<ImageMapBlock, MachineInstruction[]>();
+            Dictionary<ImageMapBlock, MachineInstruction[]> instructions = new Dictionary<ImageMapBlock, MachineInstruction[]>();
             foreach (var bi in imageMap.Items.Values.OfType<ImageMapBlock>()
```
uxmal, line 173: Style: Prefer using `var instructions....` when creating an object with `new`, especially if the type of the created object is long and awkward.

uxmal (changes requested):
I had a minor stylistic quibble, otherwise: nice job hunting down this issue!

src/UserInterfaces/WindowsForms/Controls/MixedCodeDataModel.cs
blindmatrix: Good catch, I also prefer var ..., Copy-paste mistake I guess :P

uxmal:
Thanks for the contribution!

PR #1008, opened 2021-03-08
Make VS build with internal CMake if it exists

blindmatrix:


uxmal:
Thanks for this, it will save a lot of headaches for new developers!

PR #1012, opened 2021-03-12
Spring cleaning - Project version migrator

blindmatrix:
Added code to migrate Project_v4 to Project_v5 as a separate process instead of having ProjectLoader deal with both versions internally. This simplifies the (already large) ProjectLoader and also makes it clearer what was changed between project versions (I think).

Also separated the different kinds of inputs to their own fields (and removed their common base class) as they didn't seem to share any functionality, which removed some LinQ .OfType<>() filters.

### DIFF
--- a/src/Core/Serialization/ProjectVersionMigrator.cs
+++ b/src/Core/Serialization/ProjectVersionMigrator.cs
@@ -1,3 +1,23 @@
+#region License
+/*
+ * Copyright (C) 2021-2021 Sven Almgren.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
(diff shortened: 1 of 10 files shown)


src/Core/Serialization/ProjectVersionMigrator.cs
```diff
@@ -0,0 +1,49 @@
+using System;
```
uxmal, line 1: Needs a banner, you can copy it from any other source file and put your name in it.

src/Core/Serialization/ProjectVersionMigrator.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            ArchitectureName = v4.ArchitectureName,
+            PlatformName = v4.PlatformName,
+            InputFiles = v4.InputFiles.Select(i => MigrateDecompilerInput(i)).ToList(),
+            MetadataFiles = v4.MetadataFiles,
+            AssemblerFiles = v4.AssemblerFiles,
+        };
+
+        /// <summary>
+        /// Ignored fields (They existed in v4 but are unused in current code):
```
uxmal, line 21: <Summary> explains what teh method is for. Put this in a <remarks> element further down.

src/Core/Serialization/ProjectVersionMigrator.cs
blindmatrix: Thats right, sorry! Didn't notice that because VS always collapse them for me.

src/Core/Serialization/ProjectVersionMigrator.cs
blindmatrix: That's correct, but I'm not sure if my new comments are of an appropriate scope either. Feedback is much appreciated for improvements.

uxmal:
This looks good, I'll make some small changes later. Thanks for the contribution!

PR #1014, opened 2021-03-15
UserProcedure from Procedure_v1

blindmatrix:
Tests and regressions work but an extra set of eyes would be appreciated. There are more things that might be done for this, but I feel this at least is a good start.


uxmal:
I'm going to go ahead an accept this PR despite the build failure because it was working before yesterday's push that made `ProjectLoader` more sensitive to `#nullable` violations.

PR #1016, opened 2021-04-12
Extend Python API

ptomin:
- Add byte/int16/int64 reading
- Add procedure definition by signature
- Add possibility to read memory slice
e.g `program.memory[0x10003020:0x10003030].int32`
- Add possibility to get/set output file of procedure
- Add possibility to set event handlers at script files.
Avoid direct call of "on_program_loaded" script function.
Following syntax is used instead:
```
reko.on.<event-name> += <event-handler>
```


ptomin:
`MacOS` build failed. Strange.

ptomin:
I have fixed MacOS build error.

uxmal:
Nice work, Pavel!

PR #1024, opened 2021-04-23
Redirect script console output to Reko GUI

ptomin:
- Create `Output` tab.
- Redirect console output from scripts to this tab.


ptomin:
Wrong branch

PR #1025, opened 2021-04-23
Redirect script console output to Reko GUI

ptomin:
- Create `Output` tab.
- Redirect console output from scripts to this tab.


ptomin:
`CodeFactor` failed witch `CompexCode` error at `CmdLineDriver.cs`. I think we can ignore it.

uxmal:
Fast-tracking this. I will make some changes later.

PR #1040, opened 2021-05-19
Make scripting work with segmented addresses

ptomin:
- Create sample project with segmented addresses.
- Modify `RekoProgramAPI` to accept `Reko.Core.Address` instead of
integer (linear address).
- Modify Python API to allow iteration over procedures found by scanner.
- Add possibility to read segmented memory slices.


src/Scripts/Python/RekoProgramAPI.cs
```diff
@@ -160,71 +169,78 @@ public void SetUserProcedure(ulong linearAddress, string decl)
...
+        /// Returns true if there is user procedure definition at specified
+        /// address; otherwise, false
+        /// </summary>
+        private bool TryGetUserProcedure(Address addr, out UserProcedure userProc)
         {
-            var addr = Addr(linearAddress);
-            return program.User.Procedures[addr];
+            if (program.User.Procedures.TryGetValue(addr, out userProc))
+                return true;
+            var proc = program.Procedures[addr];
```
uxmal, line 241: I would have expected the out parameter `userProc` to be null here, and have the caller make the `UserProcedure`.  Also, calling `program.Procedures[addr]` will crash if no procedure exists at address `addr`; is this the intention?

src/Core/Program.cs
```diff
@@ -686,7 +686,14 @@ public UserGlobal ModifyUserGlobal(IProcessorArchitecture arch, Address address,
                 this.ImageMap.AddItemWithSize(address, item);
             else
                 this.ImageMap.AddItem(address, item);
-
+            //$BUGBUG: what about x86 segmented binaries?
```
uxmal, line 689: If `address` has a selector, then you can look for the segment in `SegmentMap`.

src/Scripts/Python/RekoProgramAPI.cs
```diff
@@ -42,97 +42,106 @@ public RekoProgramAPI(Program program)
...
+        public void SetUserGlobal(Address addr, string decl)
         {
-            var addr = Addr(linearAddress);
             var usb = new UserSignatureBuilder(program);
             var global = usb.ParseGlobalDeclaration(decl);
             var name = global?.Name;
             var dataType = global?.DataType;
             if (name is null || dataType is null)
                 throw new ArgumentException(
                     $"Failed to parse global variable declaration: '{decl}'");
```
uxmal, line 117: Please add a '.' to the end of exception messages, for consistency.

src/Scripts/Python/RekoProgramAPI.cs
```diff
@@ -160,71 +169,78 @@ public void SetUserProcedure(ulong linearAddress, string decl)
...
             while (rdr.Offset - offStart < length)
             {
                 yield return reader(rdr);
             }
         }
 
-        private UserProcedure UserProcedure(ulong linearAddress)
+        private UserProcedure UserProcedure(Address addr)
+        {
+            TryGetUserProcedure(addr, out var userProc);
```
uxmal, line 220: Isn't this a bit of a lie? True, we have a userproc, but it's not registered in the `program`.

uxmal (commented):
I've added some comments.

src/Scripts/Python/RekoProgramAPI.cs
ptomin: Done in 608181f

src/Scripts/Python/RekoProgramAPI.cs
ptomin: The out parameter userProc to be null after 2408f40.
> calling program.Procedures[addr] will crash if no procedure exists at address addr; is this the intention

Yes, this is the intention. I have done explicit throwing of exception at  2408f40 to make it clear.

src/Scripts/Python/RekoProgramAPI.cs
ptomin: Agree. I tried to make it clear at 2408f40

src/Core/Program.cs
ptomin: I tried to add it to `ImageSegment.Fields` but it is not shown at output file. Is there support for user-defined segmented global variables?

uxmal: We should add unit tests to confirm this, both at the assembler output stage and during type inference.

ptomin: Should I attach such tests to this PR?

ptomin: I can try to create test at `https://github.com/uxmal/reko/blob/7efb910c1420c51de9d0364464cd87601a8a8554/src/UnitTests/Core/Output/SegmentFilePolicyTests.cs`.

ptomin: I have added test (failed) for `SegmentFilePolicy` at 78f9de3

ptomin: Finally, I have found solution. The fix is coming.

ptomin: Done in 162c14b and 8bc0331

ptomin:
John, I have done requested changes. Could it be merged?

uxmal:
Nice work, Pavel.

PR #1041, opened 2021-05-22
Add separate driver for Wix objects

ptomin:
- Add both WindowsDecompiler and CmdLine to new driver(WixRuntime)
- Use WixRuntime instead of WindowsDecompiler when running
"create_msi_wix" target
- Remove *.pdb and *.deps.json after build.


uxmal:
Before we go too far down this path, I'd like to have a discussion on some topics. I'm collecting my thoughts but one thing that comes immediately to mind is that neither `RunTime` nor the `WixInstallers` are "drivers" per set. Instead, they really should be categorized under "Installers". I will create a separate branch where I work on refactoring this.

Second, it's unclear why we need a new project for building the WixInstaller when there already exists one for that purpose. We need to have a discussion on how best proceed here.

ptomin:
Yes, `WixInstallers` is not driver actually. But it similar to `Runtime`, so I created it at the same folder. I agree that they should be moved to` Installers`
 >it's unclear why we need a new project for building the WixInstaller

Yes, we can try to merge `WixInstaller` and `WixRuntime`

uxmal:
I think we need to take a step back on how the installer scripts are generated. The intent is to save on labor by allowing the inclusion of the output in the `WindowsDecompiler` output directory as a source of the files that get packaged in the installer / NuGet medium. However as @ptomin discovered this doesn't work because the WiX installer is including files from projects other than `WindowsDecompiler`. 

I propose we go back to the previous "whitelist" mechanism I had in place. It was a manual step, but @smx-smx work on the installer script genenerator will make it possible to automate it. We eat an XML file, which contains a file-by-file "bill of materials" (BOM) of what is needed for each different installer/NuGet. We get the benefits of tight control of what files are included in what installer, and the benefit of having that control in one single file.

I'm a little swamped right now with other issues, but I definitely want to make headway on this so we can release a 0.9.4 / 0.10.0 version. Please have patience while I get this working right.

uxmal:
@ptomin :  Thanks for taking the time to submit this. However, after 678825c4f9ba47fe4a77709989fd12899c8533ea, the build now uses the same whitelist ($REKO/src/Installers/NuGetPackage/reko-files.xml) that it did before, so the PR has become redundant.

ptomin:
Agree. Thanks for fixing that. I'll try to test new installers and test how it works with `Reko` scripting.

PR #1046, opened 2021-06-01
Set recursion limit for Python scrips

ptomin:
- Set recursion limit for Python scrips. Avoid application crash if there is infinite recursion.
- Add script regression tests to verify raised exceptions.


src/Scripts/Python/PythonModule.cs
```diff
@@ -163,11 +163,17 @@ public override void FireEvent(ScriptEvent @event, Program program)
 
         private static ScriptEngine CreateEngine(TextWriter outputWriter)
         {
+            // Set recursion limit to avoid application crash if there is
+            // infinite recursion
+            var options = new Dictionary<string, object>()
+            {
+                { "RecursionLimit", 100 },
+            };
```
uxmal, line 171: Would it make sense to add the `options` dictionary as an optional argument to this method, so that the `RecursionLimit` could be customized? I'm seeing a hard-wired constant `100` which makes me think it might be necessary to allow it to be modified.

ptomin: I have added optional parameter in 32dc546.

PR #1053, opened 2021-06-21
Xbe fixes

smx-smx:
Fixed incorrect logic in XBE Loader

### DIFF
--- a/src/ImageLoaders/Xbe/XbeLoader.cs
+++ b/src/ImageLoaders/Xbe/XbeLoader.cs
@@ -59,7 +59,7 @@ namespace Reko.ImageLoaders.Xbe
                 return XbeBuildType.SegaChihiro;
             }
 
-            if ((ImageHeader.EntryPoint ^ XBE_XORKEY_ENTRYPOINT_RELEASE) > ImageHeader.BaseAddress)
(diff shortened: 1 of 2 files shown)


src/ImageLoaders/Xbe/XbeLoader.cs
```diff
@@ -59,7 +59,7 @@ private XbeBuildType DetectBuildType()
                 return XbeBuildType.SegaChihiro;
             }
 
-            if ((ImageHeader.EntryPoint ^ XBE_XORKEY_ENTRYPOINT_RELEASE) > ImageHeader.BaseAddress)
+            if ((ImageHeader.EntryPoint ^ XBE_XORKEY_ENTRYPOINT_RELEASE) > 0x1000000)
```
uxmal, line 62: Could you make the `0x1000000` a named constant, explaining its purpose? Remember the Reko convention is 
```C#
public const uint MagicConstant = 0x1000000
```

uxmal (commented):
Thanks for the PR. One small nit and we're good to go.

src/ImageLoaders/Xbe/XbeLoader.cs
smx-smx: Done

uxmal:
Thanks for the contribution!

PR #1055, opened 2021-07-01
Debug: Added JIT Debugger Launcher

smx-smx:
It's much faster to use when placing strategic breakpoints
This means there is less overhead (no debugger attached) while running
until we reach the point where we need to break and check


uxmal:
Thanks for the contribution!

PR #1057, opened 2021-07-10
Give full path to solution file in README

SamB:
I thought I was going crazy because I couldn't find it in the project root.


uxmal:
Thanks for the contribution!

PR #1058, opened 2021-08-02
subjects: fixed resources path having backslashes

smx-smx:


uxmal:
Thanks for taking the time to fix this.

PR #1062, opened 2021-08-13
Added support for pascal and stdcall calling conventions for MsdosPlatform

blindmatrix:
Added support for pascal and stdcall calling conventions for MsdosPlatform.

Also made underscore-prefixs optional for calling conventions for Win32Platform ("__pascal" should be treated the same as "pascal")


uxmal:
Looks good. Thanks for the PR!

PR #1063, opened 2021-08-14
Installer: add System.CodeDom.dll

ptomin:


uxmal:
Thanks for taking care of this.

ptomin:
But build is failed! It looks that there is no `System.CodeDom.dll` at `WindowsDecompiler.exe` folder. I should use `decompiler.exe` folder to obtain it.

PR #1091, opened 2021-11-18
Core.csproj: ignore exit code for 'git rev-parse'

ptomin:
Make it build for users who have not added git to PATH


uxmal:
@ptomin, could you please run the command:
```
decompile --version
``` 
and add a comment with the resulting output when `git` is not on your path. On my machine, where `git` _is_ on the path, I get the following result:
```
Decompile.exe version 0.10.2.0 (git:3d39b8462)
```
I'd like to make sure that if Reko is built without `git` available, that the output looks reasonable, perhaps something like:
```
Decompile.exe version 0.10.2.0 (git:unknown)
```

ptomin:
Outputs
```
Decompile.exe version 0.10.2.0 (git:unknown)
```
after 1192313

uxmal:
Nice work, Pavel! Thanks for fixing this.

PR #1093, opened 2021-11-20
Emit correct real values for type references

ptomin:
- Create tests to reproduce incorrect emitting real values for type references

Expected:
```
v2 = 12.0
```
but was:
```
v2 = 4.622945017495814e+18
```

### DIFF
--- a/src/Decompiler/Evaluation/IdConstant.cs
+++ b/src/Decompiler/Evaluation/IdConstant.cs
@@ -71,10 +71,11 @@ namespace Reko.Evaluation
             if (this.pt != null)
             {
                 ctx.RemoveIdentifierUse(idDst!);
+                var pt = idDst!.DataType.ResolveAs<PrimitiveType>();
                 var cNew = src!.CloneExpression();
                 if (src.DataType.IsWord &&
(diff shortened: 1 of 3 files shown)


src/Decompiler/Evaluation/IdConstant.cs
```diff
@@ -71,10 +71,11 @@ public Expression Transform()
             if (this.pt != null)
             {
                 ctx.RemoveIdentifierUse(idDst!);
+                var pt = idDst!.DataType.ResolveAs<PrimitiveType>();
                 var cNew = src!.CloneExpression();
                 if (src.DataType.IsWord &&
-                    src is Constant cSrc && 
-                    idDst!.DataType is PrimitiveType pt &&
+                    src is Constant cSrc &&
+                    pt != null &&
```
uxmal, line 78: This looks good. Just a stylistic note: let's start using `is null` and `is not null` instead of `== null` and `!= null`. The `is (not) null` operator will never call `operator ==` if a class has it defined. Not necessary for this PR, just keep it in mind.

uxmal (approved):
Nice work, @ptomin!

PR #1094, opened 2021-11-20
Do not relocate global variable from one frame to another

ptomin:
- Create test with incorrect relocation of global variable from one frame to another

`IdentifierRelocator` creates temporary variable instead.

Expected:
```
gbl_var = 123<i32>
```
but was:
```
v2 = 123<i32>
```
- Do not relocate global variable from one frame to another

### DIFF
--- a/src/Decompiler/Scanning/IdentifierRelocator.cs
+++ b/src/Decompiler/Scanning/IdentifierRelocator.cs
@@ -97,6 +97,8 @@ namespace Reko.Scanning
 
         public Identifier VisitTemporaryStorage(TemporaryStorage tmp)
         {
+            if (tmp is GlobalStorage)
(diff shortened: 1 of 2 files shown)


src/Decompiler/Scanning/IdentifierRelocator.cs
```diff
@@ -97,6 +97,8 @@ public Identifier VisitStackLocalStorage(StackLocalStorage loc)
 
         public Identifier VisitTemporaryStorage(TemporaryStorage tmp)
         {
+            if (tmp is GlobalStorage)
```
uxmal, line 100: This makes me think `GlobalStorage` merits its own visitor. You shouldn't have had to do this.

PR #1095, opened 2021-11-22
Update stack delta if jump to procedure entry discovered

ptomin:
- Create Scanner test with jump to user procedure entry

Expected:
```
stackDelta: 12
```
but was:
```
stackDelta: 0
```
-  Scanner: update stack delta after call-ret thunk was created.
If jump was replaced with call-ret then stack delta of current procedure
should be the same as for jump destination procedure.


ptomin:
`CodeFactor` thinks that
```
        private FunctionType? GetCallSignatureAtAddress(Address addrCallInstruction)
        {
            if (!program.User.Calls.TryGetValue(addrCallInstruction, out var call))
                return null;
            return call.Signature;
        }
```
is too complex procedure and raises error!

uxmal:
Codefactor is confused I think. I'm disregarding that warning.

uxmal:
This looks good Pavel. I like the improvements on the output code.

PR #1097, opened 2021-11-23
Do not set return address if it is unknown for destination procedure when creating call-ret thunk

ptomin:


uxmal:
Nice improvements, Pavel!

Issue #1051, opened 2021-06-04
Logic for opening files, projects, and archives needs a thorough refactoring


uxmal:
There are several problems with the way the Reko opens files today. One is manifested in #1048; there are multiple levels of error handling that interfere with exceptions bubbling up from a loading error (such as invalid image errors). Furthermore, the way in which Reko handles files that it considers "archives", such as floppy disk images or Stuffit archives, is very ad hoc and needs to be improved.

After successfully loading a file, there needs to be a way for the Reko GUI to understand what kind of file was loaded. We need to distinguish between regular single files, archives (with a possible default member), and project files. One possible approach is to have image loader return `object`, forcing the upper levels of Reko to probe the file by checking: 
```C#
switch (loadedObject)
{
    case Project project: OpenProject(project); break;
    case Archive archive: BrowseArchive(archive); break;
    case Program program: OpenSingleBinary(program); break;
    default: OpenAs(loadedObject);
}
```
Another approach is to introduce an interface like `ILoadedFileVisitor` similar to how `Expressions` have their `IExpressionVisitor`, but this is a low-level implementation detail.

Here are some use cases that should illustrate what I'm after:

1. A user does File/Open and selects a binary file. The loader identifies it as a ELF binary, loads it, and returns a `Program`. The project browser populates with a `Project`, consisting of a single program.

2. A user does File/Open and selects a Reko project file. The loader identifies it as such, performs a load on the project and all its internal file references, and returns a `Project`. The project browser populates with that project, which may contain multiple files, both binaries and metadata.

3. A user does File/Open and selects a binary blob of bytes. The loader fails to recognize it, so it automatically shows the `Open As...` dialog. The user enters the necessary information for the blob. This results in a `Program` object, which is used to populate the project browser as case 1

4. A user selects File/Open and selects a C64 floppy disk image. The loader recognizes it as such and returns an `Archive` . The GUI displays an `ArchiveBrowserDialog`, which allows them to pick one (or more?) files from the archive. These files are then loaded by the image loader (potentially triggering another `ArchiveBrowserDialog`. If the user finally selects an archive entry, the entry is loaded to yield a `Project` where the `<filename>` element must be annotated with care so that the archived file can be extracted again next time the project is loaded -- I'm considering a format like `filename` [`::` `path-inside-archive` ]* to allow arbitrary nesting of archives.

Because loading of files is inherently an IO operation, we will want to express all of the use cases above as `async` methods.

[...]

PR #1098, opened 2021-11-24
Archive loading

uxmal:
This PR adds support to loading "archives", of different formats, to the loading part of Reko. It is part of the overhaul specified in #1051. The work is not completed yet, and I'm continuing to work on this branch. I'm opening this PR to have other sets of eyes  look over the code to make sure I'm not making a big mistake.

The biggest changes are the `Reko.Loading.ILoader` / `Reko.Loading.Loader` class, and the fact that binary images are no longer referred to using a file system path (a simple `string`), but instead are using something I'm calling a `RekoUri` which has great similiarities to a standard `System.Uri` except that it is restricted to the `file:` schema and supports a feature that I don't think is supported by `file:` URIs. This feature is the optional addition of **fragments**, separated by '#' symbols. These fragments make it possible to express paths inside of archives (e.g. floppy disk, tape, and ISO images, ZIP-compressed files)

When using the `ILoader` interface to auto-detect the file format of an image, which may be inside one or more nested archives, you must use a `RekoUri` to specify the location of the image's bytes. Here are some examples:
```
file:///c:/users/uxmal/reko/mybinary.exe
```
This should be identical to a standard `file:` URI for a Windows file
```
file:///home/uxmal/reko/libbinary.so
```
This should be identical to a standard `file:` URI for a file on a POSIX system
```
file:///home/uxmal/reko/mac_file.hqx#macos_dir/macos_file/rsrc
```
This refers to the resource fork (`rsrc`) of a classic macOS file, compressed using Compact Pro and then hex-encoded with binhex (resulting in the path ending in `.hqx`).
```
file:///c:/users/uxmal/backup.iso#packed+files.zip#mac.hqx#macos_file/rsrc```
```
Finally, this refers to the resource fork (`rsrc`) of a a MacOS file, compressed and hex-encoded (`.hqx`), which in turn is stored inside a ZIP file, which in turn is stored on an ISO image.

Please take a look at this and provide feedback on the direction I'm taking here. One obvious next step is to provide overloads for the `ILoader.Load*` methods to support providing a simple file system path string, for the 90% case. Another is for me to evaluate whether I can replace all `RekoUrl` occurrences with `System.Url`. 

Finally, are you aware of any widely used naming conventions for referring to files inside possibly multiple levels of archives like this? I've based the '#' fragment separator on the fragment separator used in `http/s` URLs, but that concept doesn't seem to exist for the `file:` scheme. Should I create a new, separate scheme for this?


uxmal:
I will fix those broken path issues shortly.

Amusingly, this was apparently on my mind 6 years ago: https://stackoverflow.com/questions/32134432/ The answer given there links to the Java documentation for [JarURLCollection](https://docs.oracle.com/javase/7/docs/api/java/net/JarURLConnection.html) The scheme they are using is specific to `jar:` files, though. Perhaps there is another widely used scheme to express nesting this way?

uxmal:
Following discussion on Discord, I'm going to change the URI schema to `archive:` and rename `RekoUrl` to `ImageLocation`. The string serialization of an `ImageLocation` will then be:
```
archive:c:\users\bob\reko\cdrom.iso#data/binaries.tar#libs/libtest.so
```
The 90% case will still work: you should be able to use a simple string path with the `ILoader` interface.

ptomin (changes requested):
John, I have added some comments.

src/Core/UriTools.cs
```diff
@@ -0,0 +1,242 @@
...
+                    return iTripleSlash + 3;
+                }
+                else
+                {
+                    // No windows drive letter, assume Unix absolute path.
+                    return iTripleSlash + 2;
+                }
+            }
+            // Skip the 'file:' scheme.
+            return 5;
```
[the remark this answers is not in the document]
uxmal: Yeah, it's unfortunate C# doesn't know how to optimize `FileScheme.Length` to 5....

src/Decompiler/Decompiler.cs
```diff
@@ -64,11 +64,10 @@ public Decompiler(ILoader ldr, IServiceProvider services)
         /// <summary>
         /// Main entry point of the decompiler. Loads, decompiles, and outputs the results.
         /// </summary>
-        public void Decompile(string filename, string? loaderName = null)
+        public void Decompile(RekoUri imageUri, string? loaderName = null)
```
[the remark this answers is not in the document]
uxmal: Removed.

src/Decompiler/Decompiler.cs
```diff
@@ -221,87 +186,21 @@ private void BuildImageMaps()
...
-        /// <summary>
-        /// Loads a program from a file into memory using the additional information in 
-        /// <paramref name="raw"/>. Use this to open files with insufficient or
-        /// no metadata.
-        /// </summary>
-        /// <param name="fileName">Name of the file to be loaded.</param>
-        /// <param name="raw">Extra metadata supllied by the user.</param>
-        public Program LoadRawImage(string fileName, LoadDetails raw)
+        public void Assemble(RekoUri fileUri, IAssembler asm, IPlatform platform)
         {
```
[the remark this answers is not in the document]
uxmal: Removed.

src/Core/Serialization/ProjectLoader.cs
```diff
@@ -267,41 +279,87 @@ private Program LoadProgram(string projectFilePath, DecompilerInput_v5 sInput)
             }
             else
             {
-                program = (Program?)loader.LoadImage(binAbsPath, bytes, sUser.Loader, address)
-                    ?? new Program();   // A previous save of the project was able to read the file, 
+                var image = loader.LoadBinaryImage(binUri, bytes, sUser.Loader, address);
+                switch (image)
```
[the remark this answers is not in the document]
uxmal: Noted with a //$TODO as work still is in progress.

src/Decompiler/Loading/Loader.cs
```diff
@@ -132,17 +167,20 @@ private Program PostProcessProgram(string filename, Address addrLoad, ImageLoade
                     $"Image loader {imgLoader.GetType().FullName} has set the program.NeedsScanning " +
                     "and program.NeedsSsaTransform to inconsistent values.");
 
-            program.Name = Path.GetFileName(filename);
+            program.Name = Path.GetFileName(UriTools.ParseLastFragment(imageUri));
```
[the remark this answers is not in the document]
uxmal: Done.

src/Decompiler/Loading/Loader.cs
```diff
@@ -313,43 +386,48 @@ public Program LoadRawImage(string filename, byte[] image, Address? addrLoad, Lo
...
         /// <param name="offset">The offset into the array into which the file will be loaded.</param>
         /// <returns>An array of bytes with the file contents at the specified offset.</returns>
-        public virtual byte[] LoadImageBytes(string fileName, int offset)
+        public virtual byte[] LoadImageBytes(RekoUri imageUri, int offset)
         {
+            // The initial fragment is always a file system location.
+            var fragments = UriTools.ParseUriIntoFragments(imageUri.ExtractString());
             var fsSvc = Services.RequireService<IFileSystemService>();
-            return fsSvc.ReadAllBytes(fileName);
+            var imageFile = imageUri.ToString();
```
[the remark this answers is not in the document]
uxmal: Resolved.

src/Scripts/Python/PythonModule.cs
```diff
@@ -80,7 +83,7 @@ public override void FireEvent(ScriptEvent @event, Program program)
             catch (Exception ex)
             {
                 var scriptError = CreateError(
-                    Filename,
+                    Uri.ExtractString(),
```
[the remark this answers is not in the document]
uxmal: I'm using `ImageLocation.FileSystemPath` now, since currently the code doesn't support running scripts inside of archives.
Reko would have to extract/unpack them first to a temporary file, before the Python engine could execute them.

src/Gui/ProjectFilesWatcher.cs
```diff
@@ -176,9 +181,9 @@ private void OnScriptFileChanged(Project? project, string fileName)
                 return;
             var fullPath = Path.GetFullPath(fileName);
             var scriptFile = project.ScriptFiles.
-                Where(s => Path.GetFullPath(s.Filename) == fullPath).
+                Where(s => Path.GetFullPath(s.Uri.ExtractString()) == fullPath).
```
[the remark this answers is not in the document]
uxmal: Resolved.

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -814,10 +746,10 @@ public bool Save()
                 return true;
             if (string.IsNullOrEmpty(this.ProjectFileName))
             {
+                var fragment = UriTools.ParseLastFragment(
+                    decompilerSvc.Decompiler.Project.Programs[0].Uri);
                 string newName = uiSvc.ShowSaveFileDialog(
-                    Path.ChangeExtension(
-                        decompilerSvc.Decompiler.Project.Programs[0].Filename,
-                        Project_v5.FileExtension));
+                    Path.ChangeExtension(fragment, Project_v5.FileExtension));
```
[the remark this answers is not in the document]
uxmal: Done.

src/Gui/Design/ScriptFileDesigner.cs
```diff
@@ -41,13 +42,13 @@ public override void Initialize(object obj)
         public override void DoDefaultAction()
         {
             var editorSvc = Services!.RequireService<ITextFileEditorService>();
-            editorSvc.DisplayFile(scriptFile.Filename);
+            editorSvc.DisplayFile(scriptFile.Uri.ExtractString());
```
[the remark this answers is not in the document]
uxmal: I've changed this to use `FileSystemPath`. Scripts are currently assumed to not be inside archives.

src/Environments/OS2/OS2Platform16.cs
```diff
@@ -178,11 +178,12 @@ public void WriteModuleDefinition(Program program, string path)
                 fsSvc.CreateDirectory(dir);
             using (var w = fsSvc.CreateStreamWriter(path, false, new UTF8Encoding(false)))
             {
+                var fragment = UriTools.ParseLastFragment(program.Uri);
                 w.WriteLine("; {0}", Path.GetFileName(path));
-                w.WriteLine("; Generated by decompiling {0}", Path.GetFileName(program.Filename));
+                w.WriteLine("; Generated by decompiling {0}", Path.GetFileName(fragment));
```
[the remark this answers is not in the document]
uxmal: Done.

src/Decompiler/Loading/Loader.cs
```diff
@@ -194,13 +267,13 @@ public Program LoadRawImage(string filename, byte[] image, Address? addrLoad, Lo
...
 
-            var imgLoader = CreateCustomImageLoader(Services, details.LoaderName, filename, image);
+            var imgLoader = CreateCustomImageLoader(Services, details.LoaderName, imageUri, image);
             var program = imgLoader.LoadProgram(addrLoad, arch, platform);
             if (details.EntryPoint != null && arch.TryParseAddress(details.EntryPoint.Address, out Address addrEp))
             {
                 program.EntryPoints.Add(addrEp, ImageSymbol.Procedure(arch, addrEp));
             }
-            program.Name = Path.GetFileName(filename);
+            program.Name = Path.GetFileName(UriTools.ParseLastFragment(imageUri));
```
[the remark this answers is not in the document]
uxmal: Done.

src/Decompiler/Loading/Loader.cs
```diff
@@ -155,15 +193,50 @@ private Program PostProcessProgram(string filename, Address addrLoad, ImageLoade
...
+        /// no metadata.
+        /// </summary>
+        /// <param name="fileName">Name of the file to be loaded.</param>
+        /// <param name="raw">Extra metadata supllied by the user.</param>
+        public Program LoadRawImage(RekoUri uri, LoadDetails raw)
+        {
+            raw.ArchitectureOptions ??= new Dictionary<string, object>();
+            byte[] image = this.LoadImageBytes(uri, 0);
+            var program = this.LoadRawImage(uri, image, null, raw);
+            var project = new Project();
```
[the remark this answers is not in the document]
uxmal: Removed.

src/Decompiler/Loading/Loader.cs
```diff
@@ -155,15 +193,50 @@ private Program PostProcessProgram(string filename, Address addrLoad, ImageLoade
...
+        /// <paramref name="raw"/>. Use this to decompile raw blobs of data.
+        /// </summary>
+        /// <param name="fileName">Name of the file to be loaded.</param>
+        /// <param name="raw">Extra metadata supllied by the user.</param>
+        public Program LoadRawImage(byte[] image, LoadDetails raw)
+        {
+            raw.ArchitectureOptions ??= new Dictionary<string, object>();
+            var inventedUri = new RekoUri("file:image");
+            var program = this.LoadRawImage(inventedUri, image, null, raw);
+            var project = new Project();
```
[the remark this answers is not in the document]
uxmal: Removed.

subjects/scripting/pyFuncNames.reko/pySample_text.c
```diff
@@ -39,6 +39,10 @@ void initpySample()
 	//         Address: 0x10001140
 	//         File: None
 	//         Decompile: True
+	//     empty_procedure_loaded:
```
uxmal, line 42: @ptomin : is this a regression? I'm not certain of how to interpret it.

src/Decompiler/Loading/Loader.cs
```diff
@@ -83,47 +85,80 @@ public Program AssembleExecutable(string fileName, byte[] image, IAssembler asm,
...
+                return project;
+
+            var binaryImage = LoadBinaryImage(absoluteUri, image, loaderName, addrLoad);
+            if (binaryImage is Program program)
+            {
+                project = new Project();
+                project.AddProgram(absoluteUri, program);
+                project.LoadedMetadata = program.Platform.CreateMetadata();
+                program.EnvironmentMetadata = project.LoadedMetadata;
+                project.FireScriptEvent(ScriptEvent.OnProgramLoaded);
```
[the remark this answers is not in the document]
uxmal: Fixed

src/Decompiler/Loading/Loader.cs
```diff
@@ -313,43 +386,48 @@ public Program LoadRawImage(string filename, byte[] image, Address? addrLoad, Lo
...
         /// <returns>An array of bytes with the file contents at the specified offset.</returns>
-        public virtual byte[] LoadImageBytes(string fileName, int offset)
+        public virtual byte[] LoadImageBytes(RekoUri imageUri, int offset)
         {
+            // The initial fragment is always a file system location.
+            var fragments = UriTools.ParseUriIntoFragments(imageUri.ExtractString());
             var fsSvc = Services.RequireService<IFileSystemService>();
-            return fsSvc.ReadAllBytes(fileName);
+            var imageFile = imageUri.ToString();
+            return fsSvc.ReadAllBytes(fragments[0]);
```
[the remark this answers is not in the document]
uxmal: There is a challenge here. At the point where the archive is asked to load an image, it has all the metadata available to make a `Program` at that spot. Look for example on the MacOS CompactPro loader. All the information to load a resource fork is there, and so it makes sense to immediately load it. Otherwise, we have to register a magic number (there is none) or a file extension (there is none).

I like the approach that the archive loader gets a first chance of creating a `Program`. If it cannot (say, because the archive in question is a generic container, like an ISO or ZIP file) then it should return the new `Blob` image format back to the caller. The caller can then call `ILoader.LoadBinaryImage` on the bytes of `Blob`.

uxmal:
Note: I'm aware that the Unix/MacOS branches are failing. I've installed a VM to tinker with this issue.

subjects/scripting/pyFuncNames.reko/pySample_text.c
ptomin: Yes, it is. `empty_procedure_loaded` defined at https://github.com/uxmal/reko/blob/8500b5794287d1f63e228add20422d93a8d8f393/subjects/scripting/py_func_names.py#L68
which is later than `dump_procedures` https://github.com/uxmal/reko/blob/8500b5794287d1f63e228add20422d93a8d8f393/subjects/scripting/py_func_names.py#L57
So it should not be there

src/Core/Serialization/ProjectLoader.cs
```diff
@@ -202,6 +202,8 @@ public Project LoadProject(ImageLocation projectUri, Project_v5 sp)
             project.MetadataFiles.AddRange(typelibs);
             project.ScriptFiles.AddRange(scripts);
             project.Programs.AddRange(programs);
+
+            this.project.FireScriptEvent(ScriptEvent.OnProgramLoaded);
```
ptomin, line 206: Duplicated fire of `OnProgramLoaded` event. It caused regression at https://github.com/uxmal/reko/pull/1098#discussion_r758202324

ptomin: Another one is https://github.com/uxmal/reko/blob/8500b5794287d1f63e228add20422d93a8d8f393/src/Decompiler/Loading/Loader.cs#L108

src/Core/Loading/ILoadedImage.cs
```diff
@@ -23,13 +23,37 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Reko abuses the 'file:' schema to model file paths within (possibly nested) archives.
+        /// This is accomplished by using the '#' character to separate path segments within the respective 
+        /// archives.
+        /// The syntax is as follows:
+        /// - A simple executable file test.dll in the folder c:\myprograms:
+        ///     file:///c:/myprograms/test.dll
+        /// - An executable file test.so inside a TAR archive:
+        ///     file:///home/username/archive.tar#lib/test.so
```
ptomin, line 43: This documentation should be update after `archive:` syntax introducing

src/Decompiler/Decompiler.cs
```diff
@@ -64,11 +64,10 @@ public Decompiler(ILoader ldr, IServiceProvider services)
         /// <summary>
         /// Main entry point of the decompiler. Loads, decompiles, and outputs the results.
```
ptomin, line 65: Method do not load file more. Document need to be updated.

src/Decompiler/Decompiler.cs
```diff
@@ -668,14 +547,15 @@ private void WriteEntryPoints(Program program)
...
+            var irPath = Path.Combine(program.SourceDirectory, Path.GetFileNameWithoutExtension(program.Location.FilesystemPath));
             program.Platform.WriteMetadata(program, irPath);
         }
 
 		public void WriteHeaderComment(string filename, Program program, TextWriter w)
 		{
+            var fragment = program.Location.FilesystemPath;
 			w.WriteLine("// {0}", filename);
-			w.WriteLine("// Generated by decompiling {0}", Path.GetFileName(program.Filename));
+			w.WriteLine("// Generated by decompiling {0}", Path.GetFileName(fragment));
```
ptomin, line 558: Looks like it should be `program.Location.GetFilename()`

src/Decompiler/Loading/Loader.cs
```diff
@@ -83,47 +86,83 @@ public Program AssembleExecutable(string fileName, byte[] image, IAssembler asm,
...
+        /// is encountered, return that.
         /// </summary>
-        /// <param name="rawBytes">Image of the executable file.</param>
+        /// <param name="imageLocation">The location of the image to load.</param>
+        /// <param name="loaderName">Optional .NET class name of a custom
+        /// image loader</param>
+        /// <param name="addrLoad">Optional address at which to load the image.
+        /// </param>
+        /// <returns>An instance of <see cref="ILoadedImage"/> if an image could be loaded, 
+        /// or null otherwise.</returns>
```
ptomin, line 99: `an instance of <see cref="Blob"/> otherwise`

src/Core/Loading/ILoader.cs
```diff
@@ -36,68 +36,86 @@ public interface ILoader
...
 
+        /// <summary>
+        /// Loads the file specified by <paramref name="imageLocation"/>. If it is a Reko project
+        /// file, loads that. If not, tries to match the file contents to one of the known
+        /// file formats, or uses an explicity provided <paramref name="loader"/>.
+        /// </summary>
+        /// <param name="imageLocation"></param>
+        /// <param name="loader"></param>
+        /// <param name="addrLoad"></param>
+        /// <returns>Either a <see cref="ILoadedImage"/> instance, or null if the file format
```
ptomin, line 47: `or an instance of <see cref="Blob"/> if the file format`

src/Core/Loading/ILoader.cs
```diff
@@ -36,68 +36,86 @@ public interface ILoader
...
+        /// wasn't recognized.</returns>
+        ILoadedImage? Load(ImageLocation imageLocation, string? loader = null, Address? addrLoad = null);
```
ptomin, line 49: Now you return `Blob` if format is unknown. So return can be not-nullable

src/Core/Serialization/ProjectLoader.cs
uxmal: Is it appropriate to say, that the `OnProgramLoaded` event should only be fired if a project has been loaded? If a user opened a `Program` directly, there hasn't been time for the user to add a script to the project, so it makes no sense to fire the event in that case.

Perhaps we should write unit tests to cover for how this is expected to work?

src/Core/Loading/ILoadedImage.cs
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
```diff
@@ -83,47 +86,83 @@ public Program AssembleExecutable(string fileName, byte[] image, IAssembler asm,
...
+                    imgLoader = CreateDefaultImageLoader(imageLocation, image);
                 }
             }
-            if (imgLoader == null)
-                return null;
+            if (imgLoader is null)
+                return new Blob(image);
 
-            if (addrLoad is null)
+            if (addrLoad is null && imgLoader is ProgramImageLoader ploader)
```
ptomin, line 152: Consider to refactor it to
```
if (imgLoader is ProgramImageLoader ploader)
{
    if (addrLoad is null)
    {
        addrLoad = ploader.PreferredBaseAddress;     //$REVIEW: Should be a configuration property.
    }
    var program = ploader.LoadProgram(addrLoad);
    return PostProcessProgram(imageLocation, addrLoad, ploader, program);
}
return imgLoader.Load(addrLoad);
```

ptomin: Thought it will be incorrect if `imgLoader` can return `Program`. Can it do it?

src/Core/Serialization/ProjectLoader.cs
ptomin: > If a user opened a Program directly, there hasn't been time for the user to add a script to the project,

Yes.

>Perhaps we should write unit tests to cover for how this is expected to work?

It would be nice.

src/Decompiler/Decompiler.cs
uxmal: Fixed.

src/Decompiler/Decompiler.cs
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
```diff
@@ -313,43 +385,46 @@ public Program LoadRawImage(string filename, byte[] image, Address? addrLoad, Lo
...
+        /// <param name="imageLocation">Location of the image whose bytes are to be loaded.</param>
         /// <param name="offset">The offset into the array into which the file will be loaded.</param>
         /// <returns>An array of bytes with the file contents at the specified offset.</returns>
-        public virtual byte[] LoadImageBytes(string fileName, int offset)
+        public virtual byte[] LoadImageBytes(ImageLocation imageLocation, int offset)
         {
+            // The initial fragment is always a file system location.
             var fsSvc = Services.RequireService<IFileSystemService>();
-            return fsSvc.ReadAllBytes(fileName);
+            return fsSvc.ReadAllBytes(imageLocation.FilesystemPath);
```
ptomin, line 416: If we always ignore archive entry fragments what the reason of change for `fileName` to `ImageLocation`. Maybe it would be better to leave it as `public virtual byte[] LoadImageBytes(string fileName, int offset)`?

src/Core/Loading/ILoader.cs
uxmal: Fixed.

src/Core/Loading/ILoader.cs
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
uxmal: Hm, something got lost along the way here. The intent was to "drill into" the binary, expanding archives along the way. Perhaps this method no longer needs to exist. Loading a file either is recognized as an executable format, an archive format, or a blob. If it is an executable format, we're done. otherwise each archive format is responsible for extracting the bytes in whatever way is correct.

I'll review the codebase -- it's likely the `LoadImageBytes` method may disappear.

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -422,59 +391,22 @@ public bool AssembleFile()
...
-                    ArchitectureName = archName,
-                    ArchitectureOptions = dlg.ArchitectureOptions,
-                    PlatformName = envName,
-                    LoadAddress = sAddr,
-                    EntryPoint = entry,
-                };
-
-                OpenBinary(dlg.FileName.Text, (f) => pageInitial.OpenBinaryAs(f, details), f => false);
+                LoadDetails details = dlg.GetLoadDetails();
+                CloseProject();
```
ptomin, line 401: This note is unrelated to this PR but it looks like `CloseProject` can be cancelled.

ptomin (commented):
More review comments, John.

src/Core/Loading/ILoader.cs
```diff
@@ -36,68 +36,90 @@ public interface ILoader
...
+        //$TODO: Change the output to Stream; the image could be so large it doesn't
+        // fit in main memory.
+        byte[] LoadImageBytes(ImageLocation filename);
 
         /// <summary>
         /// Given a executable file image in <param name="bytes">, determines which file 
         /// format the file has and delegates loading to a specific image loader.
         /// </summary>
-        /// <param name="fileName">The file name of the executable.</param>
-        /// <param name="bytes">The contents of the executable file.</param>
```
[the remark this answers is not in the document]
uxmal: Yes, that's a mistake. Fixed.

src/Core/Loading/ImageLocation.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        /// <summary>
+        /// Computes a hash code for this <see cref="ImageLocation"/>.
+        /// </summary>
+        /// <returns>A signed integer hash code.</returns>
+        public override int GetHashCode()
+        {
+            int h = Fragments.Length;
+            if (Fragments.Length > 0)
+                h = (h * 17) ^ Fragments[^1].GetHashCode();
```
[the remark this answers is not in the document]
uxmal: I'm using a "standard trick" when computing the hash, similar to some hash methods for strings, which only consider the beginning and the end of a string. It's a tradeoff between uniqueness and speed of execution. `GetHashCode` is supposed to be quite fast. If you feel strongly about it, you can always submit a PR.

src/Decompiler/Loading/Loader.cs
```diff
@@ -83,47 +86,85 @@ public Program AssembleExecutable(string fileName, byte[] image, IAssembler asm,
...
+                project.AddProgram(imageLocation, program);
+                project.LoadedMetadata = program.Platform.CreateMetadata();
+                program.EnvironmentMetadata = project.LoadedMetadata;
+                return project;
+            }
+            return binaryImage;
+        }
+
+        /// <summary>
+        /// Loads a binary image into memory, if it has a recognized file format.
```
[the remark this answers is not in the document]
uxmal: The intent is that `LoadBinaryImage` should be able to load anything except project files, and it should be able to do it with paths or not. I'm still working on the details of how this is supposed to work.

src/Gui/Design/MetadataFileDesigner.cs
```diff
@@ -41,7 +41,7 @@ public override void Initialize(object obj)
 
         public void SetTreeNodeProperties(MetadataFile mf)
         {
-            TreeNode!.Text = Path.GetFileName(mf.Filename);
+            TreeNode!.Text = Path.GetFileName(mf.Location!.GetFilename());
```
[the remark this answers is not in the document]
uxmal: Indeed! Fixed.

src/Gui/Design/ProgramDesigner.cs
```diff
@@ -80,7 +81,15 @@ public void SetTreeNodeProperties(Program program)
...
             TreeNode.ImageName = "Binary.ico";
             var sb = new StringBuilder();
-            sb.Append(program.Filename ?? "(No file name)");
+            if (program.Location is null || string.IsNullOrEmpty(program.Location.FilesystemPath))
+            {
+                sb.Append("No file name");
+            }
+            else
+            {
+                var fragment = program.Location.GetFilename();
```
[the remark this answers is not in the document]
uxmal: Agreed, the tooltip should show the full path, including fragments.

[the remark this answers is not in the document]
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
```diff
@@ -83,47 +86,85 @@ public Program AssembleExecutable(string fileName, byte[] image, IAssembler asm,
...
+        /// </summary>
+        /// <param name="imageLocation">The location of the loaded image.</param>
+        /// <param name="image">The raw bytes fetched from <paramref name="imageLocation"/>.</param>
```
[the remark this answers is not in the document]
uxmal: The intent is that the image will be fetched from the `imageLocation`. The code is not finished yet.

src/Environments/OS2/OS2Platform16.cs
```diff
@@ -178,11 +178,12 @@ public void WriteModuleDefinition(Program program, string path)
                 fsSvc.CreateDirectory(dir);
             using (var w = fsSvc.CreateStreamWriter(path, false, new UTF8Encoding(false)))
             {
+                var fragment = program.Location.GetFilename();
```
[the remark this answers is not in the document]
uxmal: Fixed.

src/Decompiler/Decompiler.cs
```diff
@@ -221,93 +197,6 @@ private void BuildImageMaps()
...
-
-        protected Project AddProgramToProject(string fileNameWithPath, Program program)
-        {
-            if (this.project == null)
-            {
-                this.project = new Project();
-            }
-            program.Filename = fileNameWithPath;
-            program.EnsureDirectoryNames(fileNameWithPath);
-            program.User.ExtractResources = true;
```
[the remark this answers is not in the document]
uxmal: Not it's not. By default, `ExtractResources` should be `true` iirc. Thanks for catching this.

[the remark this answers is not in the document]
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
uxmal: `Loader` should now be clearer. There is a `Load` method which tries to first open a file as a `Project`, and if it isn't it opens it using `LoadBinaryImage`. If there are fragments, it attempts to 'drill' through one archive for each fragment until the end of the fragments are reached.

src/Core/Serialization/ProjectLoader.cs
```diff
@@ -279,17 +279,11 @@ private Program LoadProgram(ImageLocation projectLocation, DecompilerInput_v5 sI
...
-                case Program p:
+                if (loader.Load(binLocation, null, address) is Program p)
                     return p;
-                case IArchive arch:
-                    return LoadFromArchive(binLocation, arch)
-                        ?? new Program();
-                }
                 return new Program();   // A previous save of the project was able to read the file, 
                                         // but now we can't...
+                                        //$REVIEW: this probably should throw an exception?
```
ptomin, line 286: Yes, it should.

src/Decompiler/Loading/Loader.cs
```diff
@@ -312,44 +392,47 @@ public Program LoadRawImage(string filename, byte[] image, Address? addrLoad, Lo
...
         /// of bytes, optionally at the offset <paramref>offset</paramref>. No interpretation
         /// of those bytes is done.
         /// </summary>
-        /// <param name="fileName">File to open.</param>
-        /// <param name="offset">The offset into the array into which the file will be loaded.</param>
-        /// <returns>An array of bytes with the file contents at the specified offset.</returns>
-        public virtual byte[] LoadImageBytes(string fileName, int offset)
+        /// <param name="imageLocation">Location of the image whose bytes are to be loaded.</param>
+        /// <returns>An array of bytes with the file contents.</returns>
+        public virtual byte[] LoadImageBytes(ImageLocation imageLocation)
```
ptomin, line 420: `LoadImageBytes(ImageLocation imageLocation)` -> `LoadFileBytes(string fileName)`

src/Core/Serialization/ProjectLoader.cs
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
uxmal: Discussion of `LoadImageBytes` to follow.

src/Decompiler/Loading/Loader.cs
uxmal: Fixed.

src/Decompiler/Loading/Loader.cs
```diff
@@ -83,47 +86,91 @@ public Program AssembleExecutable(string fileName, byte[] image, IAssembler asm,
...
+            var projectLoader = new ProjectLoader(this.Services, this, Services.RequireService<DecompilerEventListener>());
+            projectLoader.ProgramLoaded += (s, e) => { RunScriptOnProgramImage(e.Program, e.Program.User.OnLoadedScript); };
+            var project = projectLoader.LoadProject(imageLocation, image);
+            if (project is not null)
+            {
+                project.FireScriptEvent(ScriptEvent.OnProgramLoaded);
+                return project;
+            }
+
+            var binaryImage = LoadBinaryImage(imageLocation, image, loaderName, addrLoad);
```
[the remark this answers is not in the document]
uxmal: Fixed.

src/Decompiler/Decompiler.cs
```diff
@@ -668,14 +573,15 @@ private void WriteEntryPoints(Program program)
...
             if (program.Platform == null)
                 return;
-            var irPath = Path.Combine(program.SourceDirectory, Path.GetFileNameWithoutExtension(program.Filename));
+            var irPath = Path.Combine(program.SourceDirectory, Path.GetFileNameWithoutExtension(program.Location.FilesystemPath));
             program.Platform.WriteMetadata(program, irPath);
         }
 
 		public void WriteHeaderComment(string filename, Program program, TextWriter w)
 		{
+            var fragment = program.Location.FilesystemPath;
```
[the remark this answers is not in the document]
uxmal: Fixed.

src/Core/Loading/ILoader.cs
```diff
@@ -36,68 +36,91 @@ public interface ILoader
...
+        /// <returns>
+        /// An <see cref="ILoadedImage"/> instance. In particular, if the
+        /// file format wasn't recognized an instance of <see cref="Blob"/> is
+        /// returned.
+        /// </returns>
+        ILoadedImage Load(ImageLocation imageLocation, string? loader = null, Address? addrLoad = null);
+
         /// <summary>
         /// Opens the specified file and reads the contents of the file,
         /// starting at file offset <paramref name="offset"/>. No interpretation
```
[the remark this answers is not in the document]
uxmal: Fixed comment. It wasn't being used anywhere.

src/Decompiler/Loading/Loader.cs
uxmal: This gets a bit confusing. A `ProgramImageLoader` must always return a `Program`. An `ImageLoader` _may_ return a `Program`, but it might also return a `Blob`, or an `IArchive`. We want to handle the case when a non-`ProgramImageLoader` returns a `Program` too.

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -247,47 +244,16 @@ public IMainForm MainForm
...
-        {
-            try
-            {
-                CloseProject();
-                SwitchInteractor(InitialPageInteractor);
-                if (openAction(file) || onFailAction(file))
-                {
-                    if (file.EndsWith(Project_v5.FileExtension))
-                    {
-                        ProjectFileName = file;
```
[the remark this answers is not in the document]
uxmal: That's a mistake. Correcting....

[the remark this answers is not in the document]
uxmal: You're right. I've made `ProjectFileName` a readonly property that returns `decompilerSvc.Decompiler.Project.Location`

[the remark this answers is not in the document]
uxmal: Fixed.

uxmal:
This is looking ready to be merged into `master`. @ptomin , what do you think?

src/Decompiler/Loading/Loader.cs
ptomin: Will `PostProcessProgram` work correctly in that case? It will ignore relocations for program returned from non-`ProgramImageLoader`.

ptomin (commented):
Almost ready to merge.
- Have you tested Opening raw binaries inside archives and then saving/restoring to project file?
- Have you tested saving new projects?

src/Core/Project.cs
```diff
@@ -19,30 +19,40 @@
...
-    public class Project 
+    public class Project : ILoadedImage
     {
         public Project()
         {
             Programs = new ObservableRangeCollection<Program>();
             MetadataFiles = new ObservableRangeCollection<MetadataFile>();
             ScriptFiles = new ConcurrentObservableCollection<ScriptFile>();
             LoadedMetadata = new TypeLibrary();
+            Location = default!;
```
ptomin, line 38: We are lying here. And it caused error at https://github.com/uxmal/reko/pull/1098#discussion_r761205395. Maybe it would be better to add explicit constructor parameter in the same way as for `AbstractHierarchicalArchive`

uxmal: "What is the cost of lies?"

I've added a constructor to `Project` that accepts an `ImageLocation`. All production code will use that. The parameter-less constructor can be used by unit tests that do don't use `ILoader`.

src/Gui/Forms/MainFormInteractor.cs
```diff
@@ -559,11 +557,7 @@ public void ShowAboutBox()
             }
         }
 
-        public string ProjectFileName
-        {
-            get { return projectFileName; }
-            set { projectFileName = value; }
-        }
+        public string ProjectFileName => decompilerSvc?.Decompiler?.Project?.Location.FilesystemPath;
```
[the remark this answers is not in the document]
uxmal: I've re-introduced `ProjectFileName` for now, while thinking about the most elegant way to solve this. I no longer see crashes when saving projects after opening executables or raw binaries.

src/Decompiler/Loading/Loader.cs
```diff
@@ -219,7 +219,7 @@ private Program PostProcessProgram(ImageLocation imageLocation, Address addrLoad
         public Program LoadRawImage(ImageLocation location, LoadDetails raw)
```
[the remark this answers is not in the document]
uxmal: Implemented.

src/Gui/Forms/InitialPageInteractor.cs
```diff
@@ -192,21 +206,38 @@ public bool OpenBinaryAs(string file, LoadDetails details)
...
+                {
+                case Program program:
+                    Debug.Assert(program.Location is not null);
+                    return Project.FromSingleProgram(program);
+                case Blob blob2:
+                    //$TODO: make 'filename' textbox readonly when there are fragments in path.
+                    var loadDetails = ShowOpenBinaryAsDialog(blob2.Location.FilesystemPath);
+                    if (loadDetails is null)
+                        return null;
+                    var rawProgram = loader.LoadRawImage(blob2.Location, loadDetails);
```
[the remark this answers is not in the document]
uxmal: Implemented.

src/Gui/Forms/MainFormInteractor.cs
uxmal: Yes, we need to do a separate pass to look at how to reach to cancelling the Save dialog.

src/Decompiler/Loading/Loader.cs
```diff
@@ -298,6 +309,30 @@ public Program LoadRawImage(ImageLocation imageLocation, byte[] image, Address?
...
+            var fsLocation = ImageLocation.FromUri(imageLocation.FilesystemPath);
+            var binaryImage = LoadBinaryImage(fsLocation, image, null, addrLoad);
+            for (int i = 0; i < imageLocation.Fragments.Length; ++i)
+            {
+                var fragment = imageLocation.Fragments[i];
+                if (binaryImage is not IArchive archive)
+                    throw new InvalidOperationException($"Image '{imageLocation}' expects an archive or file container, but none was loaded.");
+                var item = archive[fragment];
+                if (item is not ArchivedFile file)
+                    throw new InvalidOperationException($"Fragment '{fragment}' does not refer to an archived file.");
```
ptomin, line 323: We can return `file.GetBytes()` if it is last fragment
```
if (i == imageLocation.Fragments.Length - 1)
    return file.GetBytes()
```

src/Decompiler/Loading/Loader.cs
```diff
@@ -298,6 +309,30 @@ public Program LoadRawImage(ImageLocation imageLocation, byte[] image, Address?
...
+                binaryImage = file.LoadImage(Services, addrLoad);
+                if (binaryImage is Blob blob)
+                {
+                    if (i == imageLocation.Fragments.Length - 1)
+                        return blob.Image;
```
ptomin, line 328: What about case (rare case) when file format is unknown to `Reko` but user overrides it by `Open As` dialog?
I suggest move this check before loading image from archive file: https://github.com/uxmal/reko/pull/1098/commits/02db8bf1742b68f2cfb5fa4a8f35806bdd8a6e8d#r761658787

src/Decompiler/Loading/Loader.cs
```diff
@@ -298,6 +309,30 @@ public Program LoadRawImage(ImageLocation imageLocation, byte[] image, Address?
...
+                    // Archive doesn't know this format, perhaps Reko does.
+                    binaryImage = LoadBinaryImage(blob.Location, blob.Image, null, addrLoad);
+                }
+            }
+            return image;
```
ptomin, line 333: We shouldn't reach this point duiring correct execution. Throwing exception is better than returning incorrect `image`

uxmal: OK. Done.

src/Decompiler/Loading/Loader.cs
uxmal: I'm not sure I understand these two comments. Could you perhaps illustrate with some code?

ptomin: This method should return `bytes` array from specified location. If we have `ArchivedFile` and it is last fragment then we can obtain `bytes` using `GetBytes` and it's not required to call `LoadImage`. The code will be
```
if (i == imageLocation.Fragments.Length - 1)
    return file.GetBytes()
binaryImage = file.LoadImage(Services, addrLoad);
if (binaryImage is Blob blob)
{
    // Archive doesn't know this format, perhaps Reko does.
    binaryImage = LoadBinaryImage(blob.Location, blob.Image, null, addrLoad);
}
```

uxmal: Thanks, now I understand. Changed accordingly.

uxmal:
What do you think, @ptomin ? Ready for merge?

ptomin (approved):
Great job, John!

PR #1112, opened 2021-12-26
Fix incorrect 'changed' flag after simplification of sequence

ptomin:
- Create test for Expression Simplifier: `SEQ(foo_1, foo_1 + 1 + 2)`
After simplification to `SEQ(foo_1, foo_1 + 3)` expected to be:
```
    changed
```
but was:
```
    unchanged
```


uxmal:
Looks good!

PR #1127, opened 2022-01-05
Add work around: zero extend operand before slicing

ptomin:


uxmal:
Looks good Pavel. There are some recent changes that will require a conflict resolution, otherwise let's proceed.

ptomin:
Done

PR #1140, opened 2022-01-19
Refactor: The big namespace rename of 2022

uxmal:
The project has grown big, and some folders and namespaces need to be corrected to make it easier to navigate the project.


uxmal:
Let me know if you have concerns about the big renaming.

src/UnitTests/Core/Serialization/Json/JsonProjectSerializerTests.cs
```diff
@@ -32,7 +32,7 @@
 using System.Text;
 using System.Threading.Tasks;
 
-namespace Reko.UnitTests.Analysis
+namespace Reko.UnitTests.Decompiler.Analysis
```
ptomin, line 35: Maybe should be `Reko.UnitTests.Core.Serialization.Json`

ptomin (approved):
Just one little note.

src/UnitTests/Core/Serialization/Json/JsonProjectSerializerTests.cs
uxmal: Good catch!

PR #1143, opened 2022-01-21
Load unpacked Phar Lap EXP data into memory

slartibardfast:
adds LoadImage an equivalent to UnpackImage

### DIFF
--- a/src/ImageLoaders/MzExe/PharLapExtender.cs
+++ b/src/ImageLoaders/MzExe/PharLapExtender.cs
@@ -72,7 +72,10 @@ namespace Reko.ImageLoaders.MzExe
             if ((fileHeader.flags & FlagImagePacked) != 0)
             {
                 UnpackImage(fileHeader, image);
+            } else {
+                LoadImage(fileHeader, image);
             }
+            
             var loadseg = new ImageSegment("DOSX_PROG", image, AccessMode.ReadWriteExecute);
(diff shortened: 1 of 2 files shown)


src/ImageLoaders/MzExe/PharLapExtender.cs
```diff
@@ -72,7 +72,10 @@ public override Program LoadProgram(Address? addrLoad)
             if ((fileHeader.flags & FlagImagePacked) != 0)
             {
                 UnpackImage(fileHeader, image);
+            } else {
```
uxmal, line 75: Consider adhering to the coding format in the code base: `{` opens a new line always.

uxmal (commented):
Thanks for the contribution! I've just one tiny nitpick re. formatting. Also, because you're a first time contributor, please add yourself to `AUTHORS` in $(RekoDir).

src/ImageLoaders/MzExe/PharLapExtender.cs
slartibardfast: oh cool yes, I'll fix that and the whitespace as soon as I can tonight.

Thanks very much!

slartibardfast: fixed now! thanks again

-dc

uxmal:
Thanks for the contribution!

PR #1155, opened 2022-02-10
Refactor: use generic intrinsics for X86

uxmal:
This WIP shows how intrinsic functions can be generated for the X86 architecture. The PR is provided to solicit feedback from interested parties. Does it look OK? Is it too C++-centric? C-like output can be generated at the output stage by selecting a "C" vs "C++" output language and rendering the generic arguments as suffixes. E.g.
```C++
xmm0 = __divp<real32[4]>(xmm1, xmm2)
```
could be rendered 
```C
xmm0 = __divp_r32x4(xmm1, xmm2)
```


ptomin:
Looks good.

uxmal:
I'm not hearing any complaints. Pushing this to master after some conflict fixes.

PR #1157, opened 2022-02-16
Create EscapedFrameIntervalsFinder and ComplexStackVariableTransformer

ptomin:
There are two phases. `EscapedFrameIntervalsFinder` finds intervals in the
`[[-12, -4], (struct ""str"" 0008 (0 int32 dw0000) (4 real32 r0004))]`
form.  Then `ComplexStackVariableTransformer` rewrites expressions like
`fp - <offset>` if offset is inside of one of  the intervals collected by
`EscapedFrameIntervalsFinder`.
Current implementation works only if user or environment procedure
signature is defined.

### DIFF
--- a/subjects/PE/x86/pySample/shingledPySample.reko/pySample_text.c
+++ b/subjects/PE/x86/pySample/shingledPySample.reko/pySample_text.c
@@ -7,7 +7,7 @@
 // 10001000: Register (ptr32 Eq_n) fn10001000(Stack (ptr32 Eq_n) ptrArg04, Stack (ptr32 Eq_n) ptrArg08)
 PyObject * fn10001000(PyObject * ptrArg04, PyObject * ptrArg08)
 {
-	PyObject * eax_n = PyArg_ParseTuple(ptrArg08, "ii:sum", fp - 0x04, fp - 0x08);
+	PyObject * eax_n = PyArg_ParseTuple(ptrArg08, "ii:sum", &dwLoc04, &dwLoc08);
 	if (eax_n != null)
 		return Py_BuildValue("i", dwLoc04 + dwLoc08);
 	return eax_n;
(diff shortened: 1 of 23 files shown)


subjects/scripting/pyFuncNames.reko/sum.c
```diff
@@ -7,7 +7,7 @@
 // 10001000: Register (ptr32 Eq_n) sum_wrapper(Stack (ptr32 Eq_n) ptrArg04, Stack (ptr32 Eq_n) ptrArg08)
 PyObject * sum_wrapper(PyObject * ptrArg04, PyObject * ptrArg08)
 {
-	PyObject * eax_n = PyArg_ParseTuple(ptrArg08, "ii:sum", fp - 0x04, fp - 0x08);
```
uxmal, line 10: As discussed earlier, the declarations of the two new variables can be done in a future commit.

subjects/PE/x86/pySample/shingledPySample.reko/pySample_text.c
```diff
@@ -440,8 +440,8 @@ void fn10001864()
 	{
 		GetSystemTimeAsFileTime(fp - 0x0C);
 		ui32 esi_n = GetCurrentProcessId() ^ GetCurrentThreadId() ^ GetTickCount();
-		QueryPerformanceCounter(fp - 0x14);
-		ui32 esi_n = esi_n ^ (dwLoc10 ^ dwLoc14);
+		QueryPerformanceCounter(&tLoc14);
+		ui32 esi_n = esi_n ^ (tLoc14.dw0004 ^ tLoc14);
```
uxmal, line 444: What happened here? `tLoc14` seems to have sub-structure?

uxmal (commented):
Some questions regarding the output. Code looks fine.

subjects/PE/x86/pySample/shingledPySample.reko/pySample_text.c
uxmal: Can we make an issue to track this then?

ptomin: Yes, we can.

ptomin: See #1159

PR #1158, opened 2022-02-19
Fix intrinsic procedure parameters mismatch

ptomin:
- Throw exception if there is intrinsic procedure parameters mismatch. 5 tests failed.
- Fix intrinsic procedure parameters mismatch.
`rdsc` has not parameters, `trunc`/`truncf` accept one parameter


uxmal (commented):
Seems like an obvious thing to have done. Thanks for fixing this.

PR #1160, opened 2022-02-21
Do not bind stack variables to hell node if stack offset is unknown

ptomin:
We can not assume that unknown stack offset is zero. It causes incorrect parameters binding for recursive procedures:
изображение


uxmal:
Thanks for hunting this down. There are some small divergences in the regression tests. What are your thoughts on those?

ptomin:
Do you think it is regression? They look OK for me.

uxmal:
@ptomin, I've pulled the PR and sent you a comment on discord.

PR #1163, opened 2022-03-12
Add declarations for def instructions (except procedure parameters and global variables)

ptomin:


src/tests/Typing/TycoReg00014.exp
```diff
@@ -2,6 +2,7 @@
 // Return size: 0
 void fn00100000()
 fn00100000_entry:
+	ptr32 fp
```
uxmal, line 5: Should the pseudoregister `FramePointer` be excluded?

subjects/Elf/ARM/switch/switch.reko/switch_text.c
```diff
@@ -7,6 +7,7 @@
 // 00008334: void _start(Stack int32 dwArg00, Stack (ptr32 Eq_n) ptrArg08, Stack (ptr32 Eq_n) ptrArg0C, Stack (ptr32 void) ptrArg10)
 void _start(int32 dwArg00, void (* ptrArg08)(), void (* ptrArg0C)(), void * ptrArg10)
 {
+	ptr32 fp;
```
uxmal, line 10: This `fp` is there because of the `fp + 4` on the next line, right? Hopefully that will go away when `fp + 4` is replaced with `&tArg04`

uxmal (commented):
I see a lot of `fp` leaks. What is your opinion here. Ideally, once all escaped frame accesses have been taken care of, there should be no more `fp` references in generated code.

src/tests/Typing/TycoReg00014.exp
ptomin: I think it's OK. `fp` variable is used in followed code. So it should be declared.

subjects/Elf/ARM/switch/switch.reko/switch_text.c
ptomin: Yes it should go away when `fp + 4` is replaced with `&tArg04`

ptomin:
Ideally, it should be no `fp` at all. But if they exist then we should add corresponding declarations.
I hope, future modifications of `Escape Analysis` can remove most of them.

uxmal:
We'll go with this. The changes in `analysis/in-datatypes` will also help reduce the number of "loose" `fp` references.

Issue #1173, opened 2022-05-16
Incorrect structure field offsets at windows32.xml and wsock32.xml


ptomin:
Example for `windows32.xml`
```
typedef struct _OSVERSIONINFOA {
  DWORD dwOSVersionInfoSize;
  DWORD dwMajorVersion;
  DWORD dwMinorVersion;
  DWORD dwBuildNumber;
  DWORD dwPlatformId;
  CHAR  szCSDVersion[128];
} OSVERSIONINFOA, *POSVERSIONINFOA, *LPOSVERSIONINFOA;
```
converts to 
```
    <struct name="_OSVERSIONINFOA">
      <field offset="0" name="dwOSVersionInfoSize">
        <type>DWORD</type>
      </field>
      <field offset="4" name="dwMajorVersion">
        <type>DWORD</type>
      </field>
      <field offset="8" name="dwMinorVersion">
        <type>DWORD</type>
      </field>
      <field offset="12" name="dwBuildNumber">
        <type>DWORD</type>
      </field>
      <field offset="16" name="dwPlatformId">
        <type>DWORD</type>
      </field>
      <field offset="24" name="szCSDVersion">
        <arr length="128">
          <type>CHAR</type>
        </arr>
      </field>
    </struct>
```
Offset of `szCSDVersion` should be `20` instead of `24`

Example for `wsock32.xml`
```
typedef struct fd_set {
  u_int  fd_count;
  SOCKET fd_array[FD_SETSIZE];
} fd_set, FD_SET, *PFD_SET, *LPFD_SET;
```
converts to
```
 <struct name="fd_set">
      <field offset="0" name="fd_count">
        <type>u_int</type>
      </field>
      <field offset="8" name="fd_array">
        <arr length="64">
          <type>SOCKET</type>
        </arr>
      </field>
    </struct>
```
Offset of `fd_array` should be `4` but was
 https://github.com/uxmal/reko/blob/2d8b4c5d6d5da0c2674b9c5846684ead863473a1/src/Environments/Windows/wsock32.xml#L23

Other structures can be incorrect too. I think these xml file should be regenerated from headers after corresponding fix for `c2xml` (but maybe the bug was already fixed many years ago)

ptomin:
Fixed in #1174 and #1175

PR #1174, opened 2022-05-17
Regenerate windows32.xml

ptomin:
See #1173


uxmal:
Thanks for chasing this down and fixing the problem.

Issue #1151, opened 2022-01-31
Stacks don't always grow "down"


uxmal:
Reko implicitly assumes that CPU return stacks grown "down" -- towards lower addresses. This is not always the case; HP PA-RISC and PDP-10 have stack pointers that are incremented when pushing. Fixing this will affect `StackStorage` handling in a disruptive way.

PR #1182, opened 2022-06-09
Refactor: remove StackArgumentStorage and StackLocalStorage (#1151)

uxmal:
The two classes were hard-coding the direction in which the stack grows. On most architectures, the stack grows toward lower addresses in memory, but on some (like HP PA-RISC) the stack grows toward higher addresses.


ptomin (approved):
Looks good.

PR #1193, opened 2022-09-05
Xbe fixes

smx-smx:


src/Core/DynamicLinker.cs
```diff
@@ -63,14 +63,18 @@ public DynamicLinker(Project project, Program program, DecompilerEventListener e
...
             if (program.EnvironmentMetadata.Signatures.TryGetValue(svc.Name!, out var fnc))
             {
                 return fnc;
             }
-            return svc.Signature;
+            if(svc.Signature != null)
+            {
+                return svc.Signature;
+            }
+            return new FunctionType();
```
uxmal, line 77: Consider using the following more compact syntax:
```C#
    return svc.Signature ?? new FunctionType();
```

src/Core/SegmentMap.cs
```diff
@@ -71,6 +71,17 @@ public SegmentMap(Address addrBase, params ImageSegment[] segments)
 
         public Dictionary<ushort, ImageSegment> Selectors { get; }
 
+        public EndianImageReader CreateImageReader(Address address, IProcessorArchitecture arch)
+        {
+            if(!TryFindSegment(address, out var segment))
```
uxmal, line 76: Consider using standard C# formatting: space after the `if` keyword.

src/Core/SegmentMap.cs
```diff
@@ -71,6 +71,17 @@ public SegmentMap(Address addrBase, params ImageSegment[] segments)
...
+            {
+                throw new InvalidOperationException($"Address {address} not found in program segments");
```
uxmal, line 78: This is an argument validation error and should be `ArgumentException`. Also, consider putting a period ('.') at the end of the exception message.

src/ImageLoaders/Xbe/XbeLoader.cs
```diff
@@ -294,11 +297,11 @@ public override Program LoadProgram(Address? addrLoad)
                 segments.Add(tlsSegment);
             }
 
-            var imports = LoadImports();
+            SegmentMap segmentMap = new SegmentMap(ctx.EntryPointAddress, segments.ToArray());
```
uxmal, line 300: I'm not sure `ctx.EntryPointAddress` is appropriate here. The first parameter of the `SegmentMap` ctor expects the base address of the image, which isn't guaranteed to be the same as the entry point. If you don't specify the entry point address -- only passing segments -- Reko will calculate the base address as the `min` of the `ImageSegment`s' addresses.

uxmal (commented):
A few quibbles, but this looks good.

uxmal:
Thanks for the contribution!

PR #1199, opened 2022-12-07
AEON: Add instructions/comments; minor cleanup

throwaway96:
Added `l.sll`/`l.srl` decoders, rewriters, and rewriter unit tests.

Added a comment for each 6-bit opcode (except 101xxx) to aid in searching and determining the purpose of decoders at a glance.

Deleted an old decoder (`l.lwz`) that was no longer used as it's actually `l.lbz`. Moved some other comments.


uxmal:
Thanks for the contribution. I like it how more and more instructions are falling into place.

PR #1201, opened 2022-12-07
AEON: Add some instructions and tests

throwaway96:
I took out `beqi` (and some other compare/branch instructions) for now because I haven't written tests.


uxmal:
Excellent work. I especially appreciate the added unit tests.

PR #1209, opened 2022-12-09
Build Update

smx-smx:
- Force CMake to generate Visual Studio 2022 vcxproj files when running on Windows (msys/mingw is not tested/supported yet. This fixes testing within VSCode)
- Updated Cirrus CI worker to use VS2022 image
- Moved platform handling logic (x86/x64/AnyCPU) to the correct place
- Migrated the Cirrus CI macOS worker to Apple M1 (aarch64)
- Replaced the git hash generation with a cached CMake version (regenerates AssemblyData only if the git hash changes)
- (Environments) Move the invocation of c2xml from "AfterBuild" to "AfterCompile", and pull the assemblies from the "obj" dir instead of "bin". This makes c2xml part of the build, and should fix race conditions.


uxmal:
Thanks for this contribution!

PR #1211, opened 2022-12-10
AEON: Fix movhi sequence fusion; add bn.extbz, bn.sfeq, bt.sw

throwaway96:
This fixes fusion of `bX.movhi` followed by `bt.addi`, which would cause a crash before. Tests were added covering this case and some other combinations of instruction lengths.

New instructions (with tests) were also added.


uxmal:
Thanks for jumping in and fixing the movhi fuser.

PR #1212, opened 2022-12-10
AEON: Improve instruction decoding

throwaway96:
Corrected decoding of `bt.nop` and `bt.trap`. improved display of potential operands for certain Nyi opcodes.


uxmal:
Thanks for all the work!

PR #1214, opened 2022-12-11
AEON: Add bt.rfe and SPRs

throwaway96:


uxmal:
Nice work. In particular, the `F` bit is now it its appropriate place.

PR #1215, opened 2022-12-11
AEON: Enhance mfspr/mtspr intrinsics; add bt.jalr; change bn.sfltu to bn.sfgtu

throwaway96:


src/Arch/OpenRISC/Aeon/AeonRewriter.cs
```diff
@@ -529,15 +567,19 @@ private void RewriteJr()
...
                 m.Return(0, 0);
                 return;
             }
-            //$REVIEW: could be a call, r9 seems to be used as a link register.
             m.Goto(op0);
         }
 
+        private void RewriteJalr()
+        {
+            var target = (Identifier)Op(0);
```
uxmal, line 579: The cast is not needed here. `m.Call` accepts `Expression` which is a base class of `Identifier`.

uxmal:
Nice work!

PR #1217, opened 2022-12-13
AEON: Add bn.sflesi and bg.lbs instructions

throwaway96:
Includes updates to movhi sequence fuser and tests.


uxmal:
Whittling those NYI's down....

PR #1225, opened 2022-12-15
AEON: Add bt.swst

throwaway96:
This instruction stores the value of a register (other than r0) at a given 5-bit offset (left shifted by 2 bits) from the stack pointer (r1). The mnemonic is made up.

Also added `MuStack`, which provides a mutator for memory access with an unsigned offset (optionally shifted) from r1.


uxmal:
Thanks for the contribution!

PR #1239, opened 2023-01-18
Get rid of type variables emited during early type analysis

ptomin:
Avoid collision between type variables from analysis phase and typing phase. Different equivalence classes with same number causes failure during type analysis. So replace type variables emited during Data Flow Analysis


src/Decompiler/Analysis/UsedRegisterFinder.cs
```diff
@@ -213,8 +213,7 @@ SsaIdentifier SidOf(PhiArgument a)
                 }
             }
             typeDescender.MergeDataTypes();
-            return typeDescender.TypeVariables[sidParam.Identifier]
-                .DataType;
+            return typeDescender.GetType(sidParam.Identifier);
```
uxmal, line 216: An idea here, not necessarily needed to be implemented immediately. Instead of a method `GetType` on `TypeDescender` and subclasses, consider retrieving the type variables of expressions by calling `TypeStore.GetTypeVariable`

subjects/Elf/ARM/angr-685/RTOSDemo.reko/RTOSDemo_text.c
```diff
@@ -545,20 +545,20 @@ void xQueueCRReceiveFromISR(Eq_n r0, Eq_n r1, Eq_n r2)
...
-	Eq_n r7_n = *((word32) r0 + 56);
+	word32 r7_n = *((word32) r0 + 56);
 	*((word32) r0 + 0x0C) = r3_n;
 	r3_n = r3_n;
 	if (r3_n >= r4_n)
 		r3_n = *r0;
 	if (r3_n >= r4_n)
 		*((word32) r0 + 0x0C) = r3_n;
-	*((word32) r0 + 56) = (word32) r7_n - 1;
+	*((word32) r0 + 56) = r7_n + ~0x00;
```
uxmal, line 555: This is a strange regression. Is it because `r7_n` is now a `word32`?

subjects/Elf/nanoMips/ping/ping.reko/ping_text_0000.c
```diff
@@ -251,7 +251,7 @@ void main(Eq_n r0, Eq_n r4, Eq_n r5)
 	{
 		while (r17_n != 0x00)
 		{
-			word32 r6_n = *((byte) r17_n.u0 + 4);
+			word32 r6_n = *((word16) r17_n.u0 + 4);
```
uxmal, line 254: From `byte` to `word16` is a change. Is it correct?

subjects/Elf/nanoMips/ping/ping.reko/ping_text_0000.c
```diff
@@ -710,9 +710,9 @@ void ping4_run(Eq_n r0, Eq_n r2, Eq_n r3, Eq_n r4, Eq_n r5, Eq_n r6, struct Eq_n
 				break;
 			Eq_n r20_n = *r21_n;
 			*(byte *) 0x0888 = 0x00;
-			((byte) sp_n.u0 + 0x0044)->u0 = 0x00;
-			*((byte) sp_n.u0 + 72) = 0x00;
-			*((byte) sp_n.u0 + 76) = 0x00;
+			((word16) sp_n.u0 + 0x0044)->u0 = 0x00;
+			*((word16) sp_n.u0 + 72) = 0x00;
+			*((word16) sp_n.u0 + 76) = 0x00;
```
uxmal, line 715: Same comment here. Many `byte`s have become `word16`s.

subjects/Elf/nanoMips/ping/ping.reko/ping_text_0000.c
```diff
@@ -10789,40 +10789,46 @@ Eq_n fn0040738E(Eq_n r4, Eq_n r5, Eq_n r6, ptr32 & r3Out, ptr32 & r7Out, ptr32 &
...
 {
 	ptr32 fp;
 	byte tLoc0548;
 	Eq_n tLoc04C8;
 	ptr32 r11;
 	Eq_n r16_n;
-	Eq_n r19_n = r6;
-	switch (r7)
+	word32 r19_n = r6;
+	if (r7 != (byte *) 0x01)
```
uxmal, line 10803: This is a regression. `IfCascadeToSwitchRewriter` is being fooled because r7 is no longer being compared to constants but to `(byte *)` casts of constants. It should be `byte`.

subjects/Elf/nanoMips/ping/ping.reko/ping_text_0001.c
```diff
@@ -164,7 +164,7 @@ void __truncdfsf2(Eq_n r4, Eq_n r5)
 l004102FC:
 			if (!__bit<word32,word32>(r7_n, 0x1A))
 			{
-				r8_n = (byte) r8_n.u0 + 1;
+				r8_n = (word16) r8_n.u0 + 1;
```
uxmal, line 167: More of the same `byte` to `word16` change.

subjects/Elf/x86-64/ls/ls.reko/ls_text.c
```diff
@@ -357,11 +357,11 @@ void fn00000000004028C0(char ** rsi, int32 edi, struct Eq_n * fs)
 					Eq_n rdx_n = g_t61B100;
 					rax_n->u0 = 4274276;
 					g_t61B100 = rax_n;
-					*((word64) rax_n + 8) = rdx_n;
+					*((word32) rax_n + 8) = rdx_n;
```
uxmal, line 360: Why did `rax_n` get cast to `word32` ?

subjects/Elf/x86-64/ls/ls.reko/ls_text.c
```diff
@@ -1150,16 +1150,16 @@ void fn00000000004028C0(char ** rsi, int32 edi, struct Eq_n * fs)
...
 										Eq_n rax_n = g_t61AF78;
-										if (g_ptr61AF80 < (word64) rax_n + 8)
+										if (g_ptr61AF80 < (word32) rax_n + 8)
 										{
 											_obstack_newchunk(&g_t61AF60, 8);
 											rax_n = g_t61AF78;
 										}
 										*rax_n = g_t61B018;
-										g_t61AF78 = (word64) g_t61AF78 + 8;
+										g_t61AF78 = (word32) g_t61AF78 + 8;
```
uxmal, line 1159: Again, `word32` appears instead of `word64`, why?

subjects/Elf/x86-64/ls/ls.reko/ls_text.c
```diff
@@ -7536,7 +7536,7 @@ void fn000000000040E650(word32 edx, int32 esi, struct Eq_n * rdi)
 	struct Eq_n * rax_n = &g_t61B320;
 	if (rdi != null)
 		rax_n = rdi;
-	struct Eq_n * rsi_n = rax_n + ((uint64) (sil_n >> 0x05) * 0x04) /64 56;
+	struct Eq_n * rsi_n = rax_n + ((uint64) (sil_n >> 0x05) * 0x04) /64 52;
```
uxmal, line 7539: Odd change here from 56 to 52, why?

subjects/OpenRISC/sunxi/blob.reko/blob.c
```diff
@@ -4146,13 +4146,13 @@ int32 fn0000B034(Eq_n r3, int32 * r4, union Eq_n * r5)
 		{
 			*r4 = -1;
 			r6_n = -1;
-			r3_n = (uint32) ((word64) r3 - 1);
+			r3_n = (word32) r3 - 1;
```
uxmal, line 4149: I'd rather see the `uint32` be preserved here.

subjects/PE/PPC/hello_ppc.reko/hello_ppc_text.c
```diff
@@ -1962,15 +1962,15 @@ union Eq_n * fn00402CE0(union Eq_n * r2, struct Eq_n * r3, int32 r4, struct Eq_n
 			int32 r3_n = fn004036B0(r2_n, r3, &tLoc24);
 			if (r3_n == 0x00)
 			{
-				if ((tLoc24->ptr0004 & 0x03) != 0x02)
+				if ((tLoc24 + 4 & 0x03) != 0x02)
```
uxmal, line 1965: A dereference of `tLoc24` seems to have been lost here.

subjects/PE/PPC/hello_ppc.reko/hello_ppc_text.c
```diff
@@ -1937,8 +1937,8 @@ union Eq_n * fn00402CE0(union Eq_n * r2, struct Eq_n * r3, int32 r4, struct Eq_n
 {
 	ptr32 fp;
 	struct Eq_n * v5;
-	struct Eq_n ** tLoc28;
-	struct Eq_n * tLoc24;
+	word32 tLoc28;
+	word32 tLoc24;
```
uxmal, line 1941: Seems like Reko no longer identifies `tLoc24` and `tLoc28` as pointers to things.

subjects/PE/m68k/hello_m68k_CRTHEAP.c
```diff
@@ -639,11 +639,11 @@ int32 fn00002BB4(struct Eq_n * a5, Eq_n dwArg04, up32 dwArg08, struct Eq_n & a5O
 					if (d0_n - a5->dwFFFFFAD8 > 0x00)
 						a5->dwFFFFFAD8 = d0_n;
 				}
-				fn00002E18(a3_n, dwArg08, tLoc08->dw0000, &tLoc1C);
+				fn00002E18(a3_n, dwArg08, tLoc08 + 0x00, &tLoc1C);
 				struct Eq_n * a0_n = tLoc08;
-				word32 d0_n = tLoc08->dw0000 + 4;
+				word32 d0_n = tLoc08 + 0x00 + 4;
```
uxmal, line 644: In this regression we've lost a dereference.

subjects/PE/x86/RussianText/RussianText.reko/RussianText_text.c
```diff
@@ -78,6 +78,6 @@ void fn004011FC(struct Eq_n * dwArg04)
 {
 	g_dw4020F8 = (char *) fn00401158() + 28;
 	dwArg04->dw0000 = 0x82727349;
-	dwArg04->ptr0004 = &g_v4020E4;
+	dwArg04->dw0004 = 0x004020E4;
```
uxmal, line 81: This is a regression; we no longer treat 0x00402E4 as a pointer.

uxmal (commented):
This is a promising start. A lot of the `union`s we are seeing have disappeared. However, there are some strange regressions I'd like you to investigate. In particular two of the binaries have places where types have changed in size: `byte`s have become `word16` in the nanomips binary, and `word64` has become `word32` in the x64 ELF `ls` file.  There's also a few places where variables which previously were considered pointers have now lost their "pointerness" and are now just words. We need to understand the cause of those errors.

src/Decompiler/Analysis/UsedRegisterFinder.cs
ptomin: I can't get it. Do you mean refactor  `LocalTypeDescender` to remove `TypeVariables` map? `GetType` method is still needed in this case

uxmal:
Nice work Pavel. Moving forward, it may make sense to move the `TypeStore` out of the `Program` object, and instead create such objects when the analysis needs it -- once for determining local variable sizes and once for the global type analysis.

PR #1244, opened 2023-02-02
AEON: Add bn.nop

throwaway96:


uxmal:
Thanks for this contribution!

PR #1246, opened 2023-02-19
Do not reset enclosing pointer when creating array access

ptomin:
- Create unit test to reproduce incorrect build of access of array elem…ent of pointer type

Expected: `a[2<i32>]`
But was: `a[2<i32>].dw0000`

- Do not reset enclosing pointer when creating array access


uxmal:
Thanks for looking into this. I don't see any reason not to accept the PR, all regressions are improvements IMHO.

PR #1247, opened 2023-02-21
Github Actions: fix macOS build break

smx-smx:


uxmal:
Thanks for fixing this @smx-smx!

PR #1248, opened 2023-02-23
TypedExpressionRewiter: do not lose dereference of binary expression

ptomin:


uxmal:
Strange that we hadn't noticed this until now. Nice catch!

PR #1254, opened 2023-03-25
Fix clone of EnumType

ptomin:
Clone of EnumType was incorrect. Size was not copied. But making it immutable is better than adding Size to the to list of copied attributes


uxmal:
Looks good Pavel

PR #1258, opened 2023-04-14
Clone user defined signatures

ptomin:
- Create new VC sample project with user-defined procedure signatures
- Clone user-defined ot platform-defined signatures so that they can not be changed during Type Analysis. But share incomplete(ptrXX, wordXX) and unknown types so that they can be reconstructed during Type Analysis. See improvements of `PE/x86/UserDefinedSignatures` output


uxmal:
Nice improvements, Pavel!

PR #1263, opened 2023-05-10
Include C common escape characters when finding strings

gbody:
When finding strings from the dialog, wasn't including strings with common C escape characters.
common C escape characters are now considered valid.
Eg. 0x07 - 0x0d
/a 0x07, /b 0x08, /t 0x09, /n 0x0a, /v 0x0b, /f 0x0c, /r 0x0d


gbody:
@uxmal ready for review.

uxmal:
Looks good! I appreciate the addition of the unit test.

PR #1266, opened 2023-06-28
Feature: support x86 string intrinsics for segmented pointers.

uxmal:
This is a first take at making x86 string intrinsics available for real mode programs. There is one difference from the real mode implementation. When `cx` has the value 0, the `memcpy` and `memcmp` intrinsics should receive a value of 0x10000, instead of 0. I've solved this by injecting a `cx == 0  ? 0x10000 : cx` for the `size`. What do you think of the results? If this is too ugly I can abandon the effort.


ptomin (approved):
Looks good.

uxmal:
Merging. Thanks for the feedback!

PR #1268, opened 2023-07-16
Add ((id+/-e)+/-id) simplification rule

ptomin:


subjects/Elf/AVR32/ls.reko/ls_seg00001000_0000.c
```diff
@@ -17161,7 +17161,7 @@ byte fn0000EB48(Eq_n r0, Eq_n r1, ui32 r2, struct Eq_n * r3, Eq_n r4, Eq_n r5, s
 					if (r2_n <= r5_n)
 					{
 						r2_n.u0 = r5_n.u0 + 1;
-						if (r5_n.u0 + 1 < r5_n)
+						if (false)
```
uxmal, line 17164: This looks like a saturated add: first check for unsigned overflow, and if so, force the value to 0xFFFFFFFF. This regression cannot be allowed.

subjects/PE/m68k/hello_m68k_CRTHEAP.c
```diff
@@ -193,7 +193,7 @@ int32 fn0000275C(struct Eq_n * a0, struct Eq_n * a5, struct Eq_n & a0Out)
 		a5->ptrFFFFFAD0 = a0;
 		struct Eq_n * d0_n = a0;
 		struct Eq_n * d1_n = (char *) &a0->t0004 + 4;
-		if (&a0->dw0FF8 > a0)
+		if (true)
```
uxmal, line 196: This is also a regression that needs addressing.

uxmal (commented):
There are some regressions, which need to be addressed.

src/Decompiler/Evaluation/ExpressionSimplifier.cs
```diff
@@ -696,6 +696,22 @@ public virtual (Expression, bool) VisitBinaryExpression(BinaryExpression binExp)
...
+                idLeftLeft == idRight)
+            {
+                binExp = new BinaryExpression(
+                    binLeft.Operator, binExp.DataType,
+                    new BinaryExpression(
+                        binExp.Operator, binLeft.DataType,
+                        binLeft.Left, binExp.Right),
+                    binLeft.Right);
+                return (binExp, true);
+            }
```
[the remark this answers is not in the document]
ptomin: I think, (id + e) - id is always e. Given id=0xffffffff , e = 1, (0xffffffff + 1) = 0, (0 - 0xffffffff) = 1. But I need to do more detailed testing on mentioned samples. Maybe there are problems at other rules.

PR #1269, opened 2023-07-16
Fix 'x * 1 to x' simplification (binExp.Left -> left)

ptomin:
We should use simplificated left operand (left) instead of original one (binExp.Left)


uxmal:
Nice catch!

PR #1276, opened 2023-08-01
Add new sample to PE/x86/SmallBinaries (strcpy(12340034, ecx))

ptomin:
- Add new sample to PE/x86/SmallBinaries (strcpy(12340034, ecx))
- Add `((e+/-id)+/-id)` simplification rule in addition to `((id+/-e)+/-id)`


uxmal:
Thamks for this, Pavel.

PR #1282, opened 2023-09-02
Standardize minimum CMake version

throwaway96:
One CMake file had a minimum version of 3.0, while (most of) the rest had 3.10. I was getting a warning that versions prior to 3.5 would be deprecated, so I changed that one to 3.10. I also made the formatting of the `cmake_minimum_required()` lines consistent with each other and their surroundings.


uxmal:
Thanks for the PR!

PR #1283, opened 2023-09-03
Use TraceSwitch for RenderVisualization() debug logging

throwaway96:
Logging in `RenderVisualization()` is now controlled by `TraceSwitch` for `VisualizerControl`.


uxmal:
Thanks. This reduces debug spam when not needed.

PR #1284, opened 2023-09-03
Improve logging in TrashedRegisterFinder.ProcessBlock

throwaway96:
Previously, nothing would appear in the event log for certain errors that occurred in `TrashedRegisterFinder.ProcessBlock()`. Now, for any exception that is not handled elsewhere, a reference to the statement that caused the exception will be added to the log, and the containing block will be dumped.

Code by @uxmal.


uxmal:
Thanks for the PR. I will probably make this a general tool for usage in the other transformations in the analysis stage. It will help with debugging.

PR #1288, opened 2023-09-06
AEON: Make bg.beqi operand signed

throwaway96:
`bg.beqi` appears to have a 5-bit signed immediate operand, analogous to `bn.beqi`'s 3-bit operand.


uxmal:
Thanks for the contribution!

PR #1293, opened 2023-10-16
CallRewriter: copy StackDelta/FpuStackDelta when signature making

ptomin:
Do not lose stack delta after building of signature by `CallRewriter`. It's not affect output source code (only *.dis files) though. Anyway it would be nice to have consistent procedure signature data.


uxmal:
Thanks for the PR, Pavel. It's amazing we hadn't noticed this until now.

PR #1296, opened 2023-11-29
Make ComplexExpressionBuilder work correctly for dereferenced expressions with zero index and offset

ptomin:
- Add tests for ComplexExpressionBuilder(Struct array field)
```
Expected: "id->a0004"
But was:  "&id->a0004->b0000"
```
- Make ComplexExpressionBuilder work correctly for dereferenced expressions with zero index and offset.


uxmal:
Thanks for the contribution!

PR #1300, opened 2023-12-02
TypedExpressionRewiterTests: create test with multidimensional array

ptomin:
-TypedExpressionRewiterTests: create test with multidimensional array
Expected:
`r0 = (a + r1)->a0000[r2]`
But was:
`r0 = (a + r1 * 0x20<32>)[r2]`

### DIFF
--- a/src/UnitTests/Decompiler/Typing/TypedExpressionRewriterTests.cs
+++ b/src/UnitTests/Decompiler/Typing/TypedExpressionRewriterTests.cs
@@ -1567,5 +1567,46 @@ main_exit:
             RunStringTest(pb.BuildProgram(), sExp);
         }
 
+        [Test]
+        public void TerMultiDimensionalArray()
+        {
+            var sExp =
+            #region Expected
+@"// Before ///////
+// main
+// Return size: 0
+define main
+main_entry:
+	// succ:  l1
+l1:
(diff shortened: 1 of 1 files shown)


src/Decompiler/Typing/ExpressionTypeDescender.cs
```diff
@@ -216,7 +218,19 @@ public bool VisitArrayAccess(ArrayAccess acc, TypeVariable tv)
             TypeVariable tvField)
         {
             var dtElement = factory.CreateStructureType(null, elementSize);
-            dtElement.Fields.Add(0, tvField);
+            if (
+                tvField.DataType is StructureType strField &&
+                strField.Size == 0)
```
uxmal, line 223: LEt's add a comment explaining why this branch is not taken if `strField.Size != 0`

uxmal (commented):
Nice work. I've added a request for a comment, but otherwise it looks good. I like the improvements in `nbody_text.c`.

src/Decompiler/Typing/ExpressionTypeDescender.cs
ptomin: Done in faacfdf.

ptomin:
John, I have done requested change.

PR #1302, opened 2023-12-04
Use ArrayExpressionMatcher for scaling array index down

ptomin:
- Create test for `ComplexExpressionBuilder` with array access using SHL
Expected: "a[i].dw0000"
But was:  "a[(i << 3<8>) / 8<i32>].dw0000"
- Use ArrayExpressionMatcher for scaling array index down
`(i << 3<8>) / 8<i32>` -> `i`


uxmal:
Nice work. Who knew `ArrayExpressionMatcher` would come in so handy?

PR #1303, opened 2023-12-04
Fix: RISC-V CSRR instruction definition

gregoral:
CSRR instructions use the rs1 source regiter.


uxmal:
Nice work; thanks for your contribution!

PR #1305, opened 2023-12-07
Scale primitive type array index down

ptomin:
`id[index * 4]` -> `id[index]` if `id` is `word32 *`.


uxmal:
You're on a roll, @ptomin! Nice work.

PR #1307, opened 2023-12-08
ExpressionTypeDescender: keep user-defined struct fields

ptomin:
- TypedExpressionRewiterTests: create test with multidimensional array(user-defined type)

Expected: `(a + r1)->array[r2]`
But was: `(a->u0 + r1 * 8<i32>)[r2]`


uxmal:
The streak keeps going!

PR #1309, opened 2023-12-29
Sample binary with 193 distinct RiscV instructions

gregoral:
- an assembelr file with 193 risc-v instructions
- an elf binary compiled with gcc-as


uxmal:
Thanks for the sample binary!

PR #1310, opened 2024-01-03
Output external demangled procedures in the header file

ptomin:
They are sorted by (DLL name; class name; function name). External declarations from system or user-defined metafiles aren't here to avoid dublication.

### DIFF
--- a/src/Core/Program.cs
+++ b/src/Core/Program.cs
@@ -381,7 +381,7 @@ namespace Reko.Core
         /// indexed by mangled import name. External declarations from system
         /// or user-defined metafiles aren't here to avoid dublication.
         /// </summary>
-        public Dictionary<string, (string, string?, ExternalProcedure)> ExternalProcedures { get; private set; }
+        public Dictionary<string, (string?, string?, ExternalProcedure)> ExternalProcedures { get; private set; }
 
         /// <summary>
         /// List of resources stored in the binary. Some executable file formats support the
@@ -832,8 +832,24 @@ namespace Reko.Core
             fields.Add(offset, dt, name);
         }
 
+        /// <summary>
+        /// Add new imported external procedure if there is not one with same
+        /// import name.
+        /// </summary>
+        /// <param name="moduleName">
+        /// Name of module of external procedure.
+        /// </param>
+        /// <param name="importName">
+        /// Mangled import name of external procedure.
+        /// </param>
+        /// <param name="callingConvention">
+        /// Calling convention of external procedure.
+        /// </param>
+        /// <param name="ep">
+        /// <see cref="ExternalProcedure" /> object
+        /// </param>
         public void EnsureExternalProcedure(
-            string moduleName, string importName, string? callingConvention,
+            string? moduleName, string importName, string? callingConvention,
             ExternalProcedure ep)
         {
             if (ExternalProcedures.ContainsKey(importName))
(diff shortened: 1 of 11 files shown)


src/Core/Program.cs
```diff
@@ -374,6 +375,8 @@ public TypeLibraryDeserializer CreateTypeLibraryDeserializer()
         /// </summary>
         public Dictionary<string, Dictionary<FunctionType, IntrinsicProcedure>> Intrinsics { get; private set; }
 
+        public Dictionary<string, (string, string?, ExternalProcedure)> ExternalProcedures { get; private set; }
```
uxmal, line 378: Please add a doc comment for this property.

src/Core/Program.cs
```diff
@@ -822,5 +825,15 @@ public void AddGlobalField(Address address, DataType dt, string name)
             }
             fields.Add(offset, dt, name);
         }
+
```
uxmal, line 828: In PE binaries, external procedures always are associated with a module name. On other platforms, like e.g. ELF, they are not. This signature should probably be changed to accept external procedures with `null` module names.

src/Decompiler/Decompiler.cs
```diff
@@ -394,7 +394,7 @@ public void WriteGlobals(Program program, string filename, TextWriter w)
             w.WriteLine();
         }
     
-        public void WriteDecompiledTypes(Program program, string headerFilename, TextWriter w)
+        public void WriteDecompiledDecls(Program program, string headerFilename, TextWriter w)
```
uxmal, line 397: Consider using the name `WriteDecompiledDeclarations` here.

src/Decompiler/Decompiler.cs
```diff
@@ -414,6 +414,37 @@ public void WriteDecompiledTypes(Program program, string headerFilename, TextWri
...
+                p => (
+                    p.Value.Item1,
+                    p.Value.Item3.EnclosingType?.ToString(),
+                    p.Value.Item3.Name)
+            ).ToList();
+            if (externalProcedures.Count == 0)
+                return;
+            w.WriteLine("// Declarations for external procedures");
+            w.WriteLine();
+            foreach (var epKv in externalProcedures)
```
uxmal, line 433: Suggestion: you can write the header of the `foreach` loop like this:
```C#
foreach (var (importName, (dllName, convention, ep)) in externalProcedures)
...
```

uxmal (commented):
Nice work @ptomin. I've added some comments. The most significant one is about handling executable files that don't have module names associated with external procedures (e.g. ELF file format).

src/Decompiler/Decompiler.cs
ptomin: Done in 636607b

src/Decompiler/Decompiler.cs
ptomin: Done in a1e0d60

src/Core/Program.cs
uxmal: Also, let's add a doc comment here as well.

src/Core/Program.cs
ptomin: Done in 82b8276

src/Core/Program.cs
ptomin: Done in 4c6da22

ptomin:
John, I have done requested changes.

uxmal:
Nice contribution, Pavel!

PR #1311, opened 2024-01-05
Unify(double, TypeReference(struct (0 double))) => TypeReference(struct (0 double)).

ptomin:
- Create Unifier test: Unify(double, TypeReference(struct (0 double))) => TypeReference(struct (0 double))

```
Expected: "STR"
But was:  "(union (real64 u1) (STR u0))"
```
- Unify `(double, TypeReference(struct (0 double))) => TypeReference(struct (0 double))`.


src/Core/Types/Unifier.cs
```diff
@@ -237,6 +237,25 @@ private bool IsCompatibleWithMemberPointer(MemberPointer mpA, DataType b, int de
             return false;
 		}
 
+        private bool CanBeMergedInto(DataType a, DataType b)
```
uxmal, line 240: Let's add a comment here like: `Returns true if the data type a can be unified into data type b`

ptomin: Done in 4626a2b.

ptomin:
John, I have added requested comment.

uxmal:
It's done.

PR #1315, opened 2024-01-13
Make pointer to global variable if constant type is primitive type with pointer domain

ptomin:
- Create test for TypedContantRewriter: primitive type with pointer domain
Expected: `&g_t100080`
But was:  `0x100080<32>`
- Make pointer to global variable if constant type is primitive type with pointer domain


uxmal:
Looks good!

PR #1316, opened 2024-01-15
Create test for simplification of SLICE(Mem0[id + 1<32>:word64], word32, 8)

ptomin:
- Create test for simplification of SLICE(Mem0[id + 1<32>:word64], word32, 8)

Expected: `Mem0[id_3 + 2<32>:word32]`
But was:  `Mem0[id_3 + 0x00000002<p32>:word32]`


uxmal:
Fixed. Thanks for providing the unit test.

Issue #1319, opened 2024-01-26
RISC-V: Missing instructions from H and Q extensions


gregoral:
There are a number of instructions, which were recently added to RISC-V specification.
Reko supports some of them, but many are currently not supported.

**NOTE:** I'm working on a PR to add support for these instructions.

Here is a list of some unsupported instructions:
```cs
public enum Reko.Arch.RiscV.Mnemonic
{
    fabs_h,
    fadd_h,
    fclass_h,
    fcvt_d_h,
    fcvt_h_d,
    fcvt_h_l,
    fcvt_h_lu,
    fcvt_h_q,
    fcvt_h_w,
    fcvt_h_wu,
    fcvt_l_h,
    fcvt_lu_h,
    fcvt_q_h,
    fcvt_s_h,
    fcvt_w_h,
    fcvt_wu_h,
    fdiv_h,
    feq_h,
    fle_h,
    flh,
    flt_h,
    fmul_h,
    fmv_h,
    fmv_h_x,
    fmv_q_x,
    fmv_x_h,
    fneg_h,
    fsgnj_h,
    fsgnjn_h,
    fsgnjx_h,
    fsh,
    fsq,
    fsqrt_h,
    fsub_h,
    hlv_du,
    hlvx_h,
    hsv_bu,
    hsv_du,
    hsv_hu,
    lr_h,
    lr_q,
    sc_h,
    sfence_vm,
    sinval_vma,
}
```

gregoral:
Changes have been merged to main branch.

PR #1320, opened 2024-01-26
RISC-V: Instructions from H and Q extensions

gregoral:
Add missing instructions from H and Q exension.
Fix bugs with decoders for existing instructions.
Updated 193.instructions test assembly and elf binary.

Applies to: #1319


uxmal:
Nice work!

PR #1321, opened 2024-01-26
AEON: Add bt.lwst instruction

throwaway96:
The encoding for `bt.lwst` is almost the same as that of `bt.swst` except that it has the LSB set. This means the value for the immediate starts at bit 1, is 4 bits long, and is shifted left by 2 bits to calculate the offset. These instructions seem to be intended to provide more efficient stack access, using only 16 bits rather than the 24 bits of `bn.sw`/`bn.lwz`.

The `bt.lwst` and `bt.swst` mnemonics are both entirely made up. I would love to know what the official mnemonics are or if these are actually just special cases of other load/store word instructions.


uxmal:
Until we know for sure what the mnemonic should be, `bt.lwst` is perfectly adequate. Thanks for the contribution.

PR #1322, opened 2024-01-27
[CMake] fix the clean target

smx-smx:


uxmal:
Looks good. Let's add this to the documentation in `~/doc/build.md`

PR #1323, opened 2024-01-28
CParser: Define __REKO_DECOMPILER__ macro

throwaway96:
This is useful for keeping headers compatible with IDA, Ghidra, etc.

CParser doesn't have `#if` support yet, so there's no point in doing anything fancy; the value of `__REKO_DECOMPILER__` is 1.

Added a unit test.


uxmal:
LGTM, as they say. Thanks for the contribution!

PR #1329, opened 2024-02-13
AEON: bn.bnei immediate is signed

throwaway96:
Checking for overflow with a comparison to 7 doesn't make much sense. This also makes `bn.bnei` match `bn.beqi`.


uxmal:
Thanks for the contribution!

PR #1349, opened 2024-05-31
fix macOS CI (github actions)

smx-smx:
see https://github.com/actions/runner-images/issues/9741


uxmal:
Thanks for fixing this

PR #1350, opened 2024-08-15
Improve handling of nul-terminated string in Globals.

galenh:
- When StringType size is calculated, save it to the underlying ArrayType.
- Re-used existing DataTypes from Globals instead of re-creating them when collecting types.
- Fix an inverted "<=" error.
- Print linear arrays on a single line.


uxmal:
Hello @galenh ; thanks for your PR.  I have some comments that need to be addressed before accepting it. Also, as a new contributor, don't forget to add yourself to the `AUTHORS` file!

_.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
+{
+   // Use IntelliSense to find out which attributes exist for C# debugging
```
uxmal, line 2: Can you explain why this file is here, and not in `.vscode`? I'm not familiar with the significance of the leading underscore.

src/Core/Output/AbsynCodeFormatter.cs
```diff
@@ -100,13 +100,15 @@ private string UnsignedRepresentation(Constant u)
             else
             {
                 var sb = new StringBuilder();
+#if USE_TILDA_IF_SMALLER
```
uxmal, line 103: This looks like a user preference, and it shouldn't be coded using `#if` / `#endif`. Instead consider opening an issue to add support for changing the formatting of unsigned numbers, controlled by user preference. There we can discuss how this would best be accomplished.

src/Core/Output/GlobalDataWriter.cs
```diff
@@ -205,19 +205,17 @@ public bool VisitArray(ArrayType at)
             fmt.Terminate();
             fmt.Indent();
             fmt.Write("{");
-            fmt.Terminate();
-            fmt.Indentation += fmt.TabSize;
+            fmt.Write(" ");
```
uxmal, line 208: You've changed the rendering of arrays so that they are rendered on a single line. What happens when an array of 1000 entries is encountered?  I prefer the current state of affairs, until Reko can prettify the output source.

In addition, you haven't changed the `GlobalDataWriterTests.cs` unit tests to match, so the CI pipeline will fail. Make sure there is a unit tests to cover this case.

There is the beginning of a code prettifier in `Core\Output\PrettyPrinter.cs` that should be able to format arrays nicely. Consider opening an issue and using that class to make data output prettier.

src/Core/Output/TypedDataDumper.cs
```diff
@@ -182,7 +182,7 @@ public void VisitString(StringType str)
                     while (rdr.TryReadByte(out byte b))
                     {
                         //$REVIEW: assumes ASCII.
-                        if (0x20 <= b && b < 0x7F)
+                        if (0x20 >= b && b < 0x7F)
```
uxmal: You've introduced a bug. This expression will print out ASCII control characters (those whose code points are < 0x20) which is not the intent of this code. Running the unit tests would have caught this mistake. Please revert this change.

src/Core/Program.cs
```diff
@@ -669,6 +669,7 @@ public uint GetDataSize(IProcessorArchitecture arch, Address addr, DataType dt)
                     return 0;
                 while (rdr.IsValid && !rdr.ReadNullCharTerminator(strDt.ElementType))
                     ;
+                strDt.Length = (int) (rdr.Address - addr);
```
uxmal, line 672: I'm a little unhappy about mutating the string data type when it's being used in what semantically is a read-only query type function. Consider looking at setting the size when the string is created.

src/Core/Program.cs
```diff
@@ -840,6 +841,29 @@ public void AddGlobalField(Address address, DataType dt, string name)
             fields.Add(offset, dt, name);
         }
 
+        public DataType? FindGlobalField(Address address)
```
uxmal, line 844: Please add a doc comment for this new method.

src/Decompiler/Typing/TypeCollector.cs
```diff
@@ -109,7 +109,11 @@ public void CollectUserGlobalVariableTypes()
             {
                 if (ud.Value.DataType != null)
                 {
-                    var dt = ud.Value.DataType.Accept(deser);
+                    var dt = program.FindGlobalField(ud.Key);
```
uxmal, line 112: Can you write a unit tests in `TypeCollectorTests.cs` that illustrates the problem this change is trying to solve?

src/Decompiler/Typing/TypeCollector.cs
```diff
@@ -109,7 +109,11 @@ public void CollectUserGlobalVariableTypes()
...
+                    if (dt == null)
```
uxmal, line 113: Consider using the `dt is null` syntax here.

src/Reko-decompiler.sln
uxmal, line 1: SLN files are notoriously hard to view the diff on. What is the nature of the changes you've made here?

src/binall.cmd
uxmal, line 1: This looks like a private cmd file you've added. What is the purpose of "binall"? Could it be accomplished using Reko's release script? What is a `LasertType\LaserImage`?

src/buildall.cmd
uxmal, line 1: I'd prefer it if this was placed in the `src/tools` directory, to avoid clutter in the `src` directory.

src/cleanall.cmd
uxmal, line 1: Have you considered using `git clean -dfX` instead of this?

uxmal (changes requested):
I've finished the review. There are some issues that must be addressed before I can accept the PR. Notably, as it stands right now neither the unit tests nor the regression tests are passing. 

Consider running:
```
msbuild /t:run_unit_tests /v:m /m /p:Platform=x64 /p:Configuration=Release|Debug
msbuild /t:run_regressions /v:m /m /p:Platform=x86 /p:Configuration=Release|Debug
```
to discover where the test failures are.

src/Core/Output/TypedDataDumper.cs
```diff
@@ -41,11 +41,99 @@ public TypedDataDumper(EndianImageReader rdr, uint cbSize, Formatter stm)
         public void VisitArray(ArrayType at)
         {
             var addrEnd = rdr.Address + cbSize;
-            for (int i = 0; at.IsUnbounded || i < at.Length; ++i)
+
+            if (at.ElementType.Domain < Domain.Composite)
```
uxmal, line 45: Relying on an enum's value in this way is fragile. Consider exposing  a `DataType.IsComposite` property _in a separate PR_` to accomplish this.

src/Core/Output/TypedDataDumper.cs
```diff
@@ -41,11 +41,99 @@ public TypedDataDumper(EndianImageReader rdr, uint cbSize, Formatter stm)
...
         {
             var addrEnd = rdr.Address + cbSize;
-            for (int i = 0; at.IsUnbounded || i < at.Length; ++i)
+
+            if (at.ElementType.Domain < Domain.Composite) 
             {
-                if (!rdr.IsValid || addrEnd <= rdr.Address)
-                    return;
-                at.ElementType.Accept(this);
+                int offset = 0;
```
uxmal, line 47: There appears to be a lot of duplication here; could this code and the code in `VisitPrimitiveType` be harmonized?

src/Core/Output/TypedDataDumper.cs
```diff
@@ -198,13 +288,20 @@ public void VisitString(StringType str)
...
                             sep = ",";
                             fmt.Write(string.Format("0x{0:X2}", b));
                             if (b == 0)
                                 break;
                         }
+                        i++;
+                        Debug.Assert(i < size);
+                    }
+                    if (inStringLiteral)
+                    {
```
uxmal, line 303: Please add a unit tests to show how this is an improvement.

uxmal (commented):
I've added a few more comments. The changes are starting to become overwhelming, and could benefit from being broken up into separate, more focused PR's

PR #1358, opened 2025-04-08
fix null reference for unknown ELF ABIs

smx-smx:
`platform` was left uninitialized for unknown ELF ABIs.
Since `GetEnvironment` already returns `DefaultPlatform` for unknown environments, we can just call it


uxmal:
Nice catch! Thanks for fixing this.

PR #1370, opened 2026-08-03
Fix MN103 decoding of (d32,SP) operands and CALLS with 32-bit displacement

sympho-ru:
While disassembling MN103 camera firmware with Reko I ran into two decoding
bugs in the FC-prefix (6-byte instruction) page of the MN103 disassembler:

**1. `d32_sp` reads only 16 of its 32 displacement bits.**

```csharp
private static readonly Mutator<MN103Disassembler> d32_sp = reg_relative_u16(Registers.sp);
```

Every `(d32,SP)` memory operand (the `MOV`/`MOVBU`/`MOVHU` forms at
`FC 8x/9x/Bx`) decoded with a truncated displacement and a length of 4 bytes
instead of 6, desynchronizing the instruction stream from that point on. For
example, `FC B4 78 56 34 12` (`mov (0x12345678,sp),d0`) decoded as
`mov (5678,sp),d0` with length 4, and the remaining displacement bytes were
then decoded as the start of the next instruction.

**2. `CALLS` with a 32-bit displacement (`FC FF`) decoded as `calls (d32,sp)`.**

This form is PC-relative — the target is the address of the `CALLS`
instruction plus the displacement — analogous to the 16-bit form (`FA FF`),
which already decodes with `d16_pc`. With the fix the operand is the resolved
call target address, so far calls no longer appear as SP-relative memory
accesses.

References:

- Panasonic MN1030/MN103S Series Instruction Manual (13250-040E): the CALLS
  specification (p. 88) gives `PC+6 → mem32(SP), PC+6 → MDR, PC+d32 → PC`,
  size 6, for the 32-bit form (this also answers the `//$REVIEW` question in
  `d32_pc`: the displacement is anchored at the start of the instruction);
  the FC-prefix instruction map (p. 189) lists `CALLS (d32,PC)` at `FC FF`
  and full 32-bit displacements for the `(d32,SP)` moves.
- GNU binutils `opcodes/m10300-opc.c` agrees:
  `"calls" 0xfcff0000 … {IMM32_PCREL}`, and the `(d32,SP)` moves are `FMT_D4`
  (6 bytes) with `IMM32`.

Added six disassembler tests covering the affected forms and two rewriter
tests (`calls (d32,PC)`, `mov (d32,SP),Dn`). All eight fail without the fix;
the whole MN103 suite passes with it.

One observation left out of scope: `calls` decoder entries carry no
`InstrClass.Transfer | InstrClass.Call`, so `calls` instructions report
`InstrClass.Linear` (visible in the existing `Mn103Rw_calls` test's `L--`
cluster class). Happy to address that in a follow-up if there's interest.

### DIFF
--- a/src/Arch/MN103/MN103Disassembler.cs
+++ b/src/Arch/MN103/MN103Disassembler.cs
@@ -243,7 +243,7 @@ namespace Reko.Arch.MN103
                 return true;
             };
         }
-        private static readonly Mutator<MN103Disassembler> d32_sp = reg_relative_u16(Registers.sp);
+        private static readonly Mutator<MN103Disassembler> d32_sp = reg_relative_u32(Registers.sp);
 
 
         private static bool d8_pc(uint uInstr, MN103Disassembler dasm)
@@ -905,7 +905,7 @@ namespace Reko.Arch.MN103
                     invalid,
                     invalid,
                     Instr(Mnemonic.add, imm32, sp),
-                    Instr(Mnemonic.calls, d32_sp)));
+                    Instr(Mnemonic.calls, d32_pc)));
             var fddecoders = Mask(2, 6, "  FD",
                 Instr(Mnemonic.udf00, imm32, dn),
                 Instr(Mnemonic.udfu00, imm32, dn),
--- a/src/UnitTests/Arch/MN103/MN103DisassemblerTests.cs
+++ b/src/UnitTests/Arch/MN103/MN103DisassemblerTests.cs
@@ -47,6 +47,52 @@ namespace Reko.UnitTests.Arch.MN103
 
         public override Address LoadAddress => addr;
 
+        private void AssertCode(string sExpected, string hexBytes)
+        {
+            var instr = DisassembleHexBytes(hexBytes);
+            Assert.AreEqual(sExpected, instr.ToString());
+        }
+
+        [Test]
+        public void MN103Dis_calls_d32()
+        {
+            // FC FF: calls (d32,PC); the target is relative to the address
+            // of the calls instruction itself.
+            AssertCode("calls\t00101234", "FCFF34120000");
+        }
+
+        [Test]
+        public void MN103Dis_calls_d32_negative_displacement()
+        {
+            AssertCode("calls\t000FFF00", "FCFF00FFFFFF");
+        }
+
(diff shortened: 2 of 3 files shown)


uxmal:
Thanks for the PR! Consider adding yourself to the `AUTHORS`. 

The `InstrClass.Call` value recently changed interpretation. Previously it was a bitflag that needed to be merged together with the `Transfer` bitflag. It was renamed to `CtiCall`, and the resulting `Transfer | CtiCall` is now the `Call` value in the `InstrClass` enumeration.

I wasn't able to find a "real" MN103 binary to test with Reko. If you have a binary you'd feel comfortable sharing with me, consider contacting me with details.

sympho-ru:
Thanks! I’m adding my name to `AUTHORS` and updating all three `CALLS` forms to use the current composite instruction classes: `InstrClass.CallInd` for `CALLS (An)`, and `InstrClass.Call` for the d16/d32 PC-relative forms, with corresponding classification tests.

The MN103 image I have been using is the Olympus PEN E-PL3 Body 1.6 firmware:
* Filename: `OLY_E_086_1600_0000_0000.BIN`
* Size: 46596160 bytes
* SHA-256: `89b70dd65c6739de1cd762777205953ee15864df73369b84815df7ed8a0eb4b1`
* Olympus model code: `E086`

OM lists E-PL3 Body 1.6 here:
https://support.jp.omsystem.com/en/support/imsg/digicamera/download/software/firm/e1/

A community browser-based route is the Olympus Firmware Updater Project:
https://ofu.infinityfreeapp.com/

This walkthrough also contains a script that requests the signed payload URL from OM’s own firmware service:
https://marcrphoto.wordpress.com/2026/05/11/olympus-om-system-firmware-updates-the-situation-changed-again-may-2026/

The firmware is a five-block Olympus container rather than an ELF. Its decoded main-code block loads at `0x402c0000`; the first 256 bytes alone contain five genuine `CALLS d32` instructions.

I’m preparing a separate public research preview covering the firmware container, disassembly workflow, experiment records and reproduction steps. The Reko MN103 decoder changes are on the critical path to publishing this. I’ll share the research repository separately when it is ready.

uxmal:
Thanks for the contribution. I'm very interested to hear more about your research results.

